PR #4, opened 2015-02-10
Provision a heroku box on the fly

commuterjoy:
Uses [codeship env variables](https://codeship.com/documentation/continuous-integration/set-environment-variables/) and the [heroku api](https://github.com/heroku/node-heroku-client) to provision an environment ready for deployment.

I've gone with `<project>-<branch>-<build-number>` as a convention.

todo - 

- Tweak deployment to use this convention, Eg. make deploy --app ft-next-engels-foo-1
- Tell the night-watch smoke test to use this convention.
- Write a teardown script to happen on branch merge/deletion.

/cc @matthew-andrews might need you advice/help with the deployment tweak.

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,9 +6,11 @@
   "version": "1.3.2",
   "dependencies": {
     "commander": "^2.6.0",
+    "debug": "^2.1.1",
     "denodeify": "^1.2.0",
     "es6-promise": "^2.0.1",
     "haikro": "^1.9.0",
+    "heroku-client": "^1.9.1",
     "isomorphic-fetch": "^1.1.0"
   }
 }
--- a/package.json
+++ b/package.json
@@ -9,7 +9,7 @@
     "debug": "^2.1.1",
     "denodeify": "^1.2.0",
     "es6-promise": "^2.0.1",
-    "haikro": "^1.9.0",
+    "haikro": "^1.12.0",
     "heroku-client": "^1.9.1",
     "isomorphic-fetch": "^1.1.0"
   }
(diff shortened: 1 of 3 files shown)


matthew-andrews:
Also hook up to next config vars to set environment variables of the app being deployed.

package.json
```diff
@@ -6,9 +6,11 @@
   "version": "1.3.2",
   "dependencies": {
     "commander": "^2.6.0",
+    "debug": "^2.1.1",
     "denodeify": "^1.2.0",
     "es6-promise": "^2.0.1",
     "haikro": "^1.9.0",
+    "heroku-client": "^1.9.1",
```
matthew-andrews: I've had pretty poor experiences using the heroku client. No promises, does weird things around pagination… generally felt like it was adding more complexity than it saved.  If you don't mind me switching your new code to just make the raw api call I'd be happy to do it…

matthew-andrews: Also as an alternative haikro already supports provision in this sense through it's create task.  That would need expanding to allow organization to be set though.

matthew-andrews: And region

commuterjoy: Ok, let me give that a go.

matthew-andrews: when [this build finishes](https://travis-ci.org/matthew-andrews/haikro/builds/50318687) you should be able to do:-

```js
var provision = require('haikro/lib/create');
create({ token: token, app: 'my-app-name', region: 'eu', organization: 'financial-times' })
  .then(function() {
    // success
  });
```

commuterjoy:
@matthew-andrews ok, I've reworked this to use your new haikro branch

package.json
```diff
@@ -9,8 +9,11 @@
   },
   "dependencies": {
     "commander": "^2.6.0",
+    "debug": "^2.1.1",
     "denodeify": "^1.2.0",
     "es6-promise": "^2.0.1",
+    "haikro": "^1.12.0",
+    "heroku-client": "^1.9.1",
```
matthew-andrews: is this unused now?

PR #5, opened 2015-02-11
Allow the configure task to target other app names

commuterjoy:
```
# configure based on package.json in cwd
node ./bin/next-build-tools configure

# configure engels
node ./bin/next-build-tools configure ft-next-engels

# configure ft-next-engels-pr-1 with ft-next-engels' configuration
node ./bin/next-build-tools configure ft-next-engels ft-next-engels-pr-1 
```

### DIFF
--- a/bin/next-build-tools.js
+++ b/bin/next-build-tools.js
@@ -34,10 +34,10 @@ program
 		});
 
 	program
-		.command('configure')
+		.command('configure [source] [target]')
 		.description('downloads environment variables from next-config-vars and uploads them to the current app')
-		.action(function() {
-			configure().catch(exit);
+		.action(function(source, target) {
+			configure({ source: source, target: target }).catch(exit);
 		});
 
 	program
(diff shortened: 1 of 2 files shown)


matthew-andrews:
:+1: yup make sense

bin/next-build-tools.js
```diff
@@ -34,10 +34,10 @@ program
 		});
 
 	program
-		.command('configure')
+		.command('configure [source] [target]')
```
matthew-andrews: ah wait

matthew-andrews: should this be `option('--source')`?  not that familiar with commander.js

matthew-andrews: no actually i  like your api better

matthew-andrews:
:+1:

bin/next-build-tools.js
commuterjoy: Yup

square brackets - optional, chevrons - required

PR #7, opened 2015-02-11
Enforce usage of next-config-vars service by breaking deploys if not used

matthew-andrews:
Up to you @commuterjoy (& team).  I had this in before but took it out because I didn't want to disrupt people.

### DIFF
--- a/bin/next-build-tools.js
+++ b/bin/next-build-tools.js
@@ -29,7 +29,11 @@ program
 		.command('deploy [app]')
 		.description('runs haikro deployment scripts with sensible defaults for Next projects')
 		.action(function(app) {
-			deploy(app).catch(exit);
+			configure({ source: app })
+				.then(function() {
+					return deploy(app);
+				})
+				.catch(exit);
 		});
 
 	program


commuterjoy:
I vote yes. But how complete/correct is the config-vars set of configs? It looks to be missing quite a few.

matthew-andrews:
So yeah it would break every build… encouraging those vars to be filled out quickly — quite disruptive though!  But good to spread the knowledge of how it works..

commuterjoy:
That's fine, but need to make sure it won't delete vars that people have added in last couple of weeks!

Maybe throw an error rather than delete vars for a week.

matthew-andrews:
hmm…  i don't mind manually checking the ones that are in config-vars (there's only a handful)

commuterjoy:
ok - perhaps i'll pick on someone to add the rest next week, then merge this.

commuterjoy:
I wonder if this is overloading deploy and that Makefile could keep these things separate...

```
next-build-tools provision ft-next-foo
next-build-tools configure ft-next-engels ft-next-foo 
next-build-tools deploy ft-next-foo
```

commuterjoy:
http://git.svc.ft.com:8080/projects/NEXTPRIVATE/repos/config-vars/browse - all env vars are now in the api

matthew-andrews:
i like the solution we currently have, separation of concerns etc.

PR #8, opened 2015-02-11
Allow config to be overridden on the fly

commuterjoy:
When copying config we want to overwrite the odd value. In this case we don't want feature branch builds behaving as production boxes. (I think) For example, do we want these branches logging to Graphite?

```
next-build-tools configure ft-next-engels ft-next-engels-4014692 \
    --overrides "NODE_ENV=branch,DEBUG=*"
```


commuterjoy:
Nb. `next-build-tools provision|configure` should probably replace `heroku create` for new apps.

matthew-andrews:
:+1:

PR #10, opened 2015-02-12
Noprune

matthew-andrews:
@commuterjoy do you mind taking over this to test & (hopefully) roll out


matthew-andrews:
might fail — may need to re run once this build finishes
https://travis-ci.org/matthew-andrews/haikro/builds/50536689

PR #15, opened 2015-03-10
next-build-tools so we don't have to have /*jshint …*/\n/*global …*/ at the top of every file

matthew-andrews:
@wheresrhys @commuterjoy @ironsidevsquincy @aintgoin2goa @adgad


config/jshint.json
```diff
@@ -0,0 +1,25 @@
+{
+  "node": true,
```
matthew-andrews: this is added

config/jshint.json
```diff
@@ -0,0 +1,25 @@
...
+  "expr": true,
+  "gcl": true,
+  "globalstrict": true,
+  "lastsemic": true,
+  "multistr": true,
+  "smarttabs": true,
+  "sub": true,
+  "browser": true,
+  "nonstandard": true,
+  "predef": ["requireText", "fetch"]
```
matthew-andrews: as is `fetch` here

config/jshint.json
```diff
@@ -0,0 +1,26 @@
+{
+  "node": true,
+  "eqeqeq": true,
+  "forin": true,
+  "freeze": true,
+  "immed": true,
+  "laxbreak": true,
```
matthew-andrews: and finally this also

PR #19, opened 2015-03-12
Destroy before provision in case a branch build runs within 10 mins of the last branch build

### LINKED ISSUE #17: Provision when app already exists
If rerunning a build, the provision task fails because the app already exists. 

Should delete/update the app.

matthew-andrews:
Fixes #17


matthew-andrews:
cc @commuterjoy & @adgad

PR #22, opened 2015-03-17
Don't exit until the app is __gtg

matthew-andrews:


matthew-andrews:
cc @commuterjoy

matthew-andrews:
this does the trick.

PR #25, opened 2015-03-19
add ios 7

matthew-andrews:
@spriggsha just made a guess here.  where are the docs for this…?

### DIFF
--- a/config/nightwatch.json
+++ b/config/nightwatch.json
@@ -95,6 +95,14 @@
         "device-orientation": "portrait"
       }
     },
+    "iphone7": {
+      "desiredCapabilities": {
+        "platform": "OS X 10.10",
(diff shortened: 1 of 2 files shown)


spriggsha:
@matthew-andrews  That should work.  Sauce config docs are at https://docs.saucelabs.com/reference/platforms-configurator/#/

config/nightwatch.json
```diff
@@ -95,6 +95,14 @@
         "device-orientation": "portrait"
       }
     },
+    "iphone7": {
+      "desiredCapabilities": {
+        "platform": "OS X 10.10",
+        "version": "7",
```
matthew-andrews: does this need to be `7.0`?

PR #26, opened 2015-03-19
Android Nexus 7HD in nw config

spriggsha:
Emulates stock browser (default on HTC and Samsung devices) which currently fails js-success test.

@matthew-andrews @commuterjoy

### DIFF
--- a/config/nightwatch.json
+++ b/config/nightwatch.json
@@ -102,6 +102,15 @@
         "deviceName": "iPad Simulator",
         "device-orientation": "portrait"
       }
+    },
+    "Android_Nexus7HD": {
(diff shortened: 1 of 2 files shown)


matthew-andrews:
do you want to add that to the default `env` config too so that it's used on each build?

spriggsha:
@matthew-andrews Yes.  If we're killing builds by supporting extra browsers, this is the way to do it.

matthew-andrews:
@spriggsha that will need updating in this repo now :)
https://github.com/Financial-Times/next-build-tools/blob/master/tasks/nightwatch.js#L13

matthew-andrews:
:+1:

PR #33, opened 2015-03-26
added task to verify bower templates used in layouts exist

wheresrhys:
@matthew-andrews


matthew-andrews:
oh that's pretty neat

PR #37, opened 2015-03-30
Assets on s3

adgad:
/cc @matthew-andrews @commuterjoy @richard-still-ft 

Task to deploy static assets to Amazon S3

### DIFF
--- a/bin/next-build-tools.js
+++ b/bin/next-build-tools.js
@@ -98,8 +98,8 @@ program
 		.description('deploys ./hashed-assets/ to <app-name> on GitHub')
 		.action(function() {
 			Promise.all([
-			  deployHashedAssets(),
-			  deployHashedAssetsToS3()
+				deployHashedAssets(),
+				deployHashedAssetsToS3()
 			]).catch(exit);
 		});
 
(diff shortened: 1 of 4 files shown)


matthew-andrews:
can we keep the same api and have the `deploy-hashed-assets` method deploy to both (initially optionally to S3 — warning but not failing if no API key set)?

adgad:
@matthew-andrews like so?

bin/next-build-tools.js
```diff
@@ -97,6 +98,7 @@ program
 		.description('deploys ./hashed-assets/ to <app-name> on GitHub')
 		.action(function() {
 			deployHashedAssets().catch(exit);
+			deployHashedAssetsToS3().catch(exit);
```
matthew-andrews:
```js
Promise.all([
  deployHashedAssets,
  deployHashedAssetsToS3
]).catch(exit);
```

package.json
```diff
@@ -9,6 +9,7 @@
     "test": "./bin/next-build-tools.js verify"
   },
   "dependencies": {
+    "aws-sdk": "^2.1.19",
```
matthew-andrews: urgh that dep is HUGE

adgad:
hmm...let me rewrite using the REST api.

adgad:
Okay I'm giving up on trying to get the REST api working for now :( Will get this through/working and can optimise later if needed.

matthew-andrews:
:+1:

PR #57, opened 2015-04-14
Adding deploy vcl task

aintgoin2goa:
So apps with their own fastly config can deploy it themselves

### DIFF
--- a/bin/next-build-tools.js
+++ b/bin/next-build-tools.js
@@ -134,7 +134,7 @@ program
 	program
 		.command('deploy-vcl [folder]')
 		.description('Deploys VCL in [folder] to the fastly service given in the FASTLY_SERVICE_ID env var.  Also requires FASTLY_KEY env var')
-		.option('m, --main <main', 'Set the name of the main vcl file (the entry point).  Defaults to "main"')
+		.option('m, --main <main', 'Set the name of the main vcl file (the entry point).  Defaults to "main.vcl"')
 		.action(function(folder, options){
 			if(folder){
 				deployVcl(folder, options);
--- a/bin/next-build-tools.js
+++ b/bin/next-build-tools.js
@@ -134,7 +134,8 @@ program
 	program
 		.command('deploy-vcl [folder]')
 		.description('Deploys VCL in [folder] to the fastly service given in the FASTLY_SERVICE_ID env var.  Also requires FASTLY_KEY env var')
-		.option('m, --main <main', 'Set the name of the main vcl file (the entry point).  Defaults to "main.vcl"')
+		.option('-m, --main <main', 'Set the name of the main vcl file (the entry point).  Defaults to "main.vcl"')
+		.option('-v, --vars <vars>', 'A way of injecting environment vars into the VCL.  So if you pass --vars AUTH_KEY,SERVICE the values {$AUTH_KEY} and ${SERVICE} in the vcl will be replaced with the values of the environmemnt variable')
 		.action(function(folder, options){
 			if(folder){
 				deployVcl(folder, options);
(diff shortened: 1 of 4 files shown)


matthew-andrews:
ohhh yes i like this.

matthew-andrews:
i'm surprised it's not complained at you for not updating the readme >_>

aintgoin2goa:
It does locally.  Nice touch :)  Will fix now

matthew-andrews:
yeah md5 doesn't seem availalble in Travis…… might ask them about that……

package.json
```diff
@@ -11,6 +11,7 @@
     "debug": "^2.1.1",
     "denodeify": "^1.2.0",
     "es6-promise": "^2.0.1",
+    "fastly": "^1.2.1",
```
matthew-andrews: think i might instigate a one in one out policy for dependencies…… don't want an Origami Build Tools 2 minute installation……

package.json
```diff
@@ -18,7 +19,8 @@
     "heroku-client": "^1.9.1",
     "isomorphic-fetch": "^2.0.0",
     "origami-build-tools": "^2.13.0",
-    "semver": "^4.3.1"
+    "semver": "^4.3.1",
+    "fastly": "git+https://github.com/Financial-Times/fastly.git#v2.5.0"
```
matthew-andrews: seems duplicated……

also

`Financial-Times/fastly#v2.5.0` cleaner :wink:

aintgoin2goa: Thanks will fix.  Also,  according to npm-check I can get rid of debug and heroku-client.  does that sound right?  I know you hate debug and I guess heroku-client is a dep of haikro anyway

matthew-andrews: i don't hate debug, i just don't use it

matthew-andrews: heroku-client isn't a dependency of haikro, haikro just talks to the REST api directly but that sounds good!

aintgoin2goa:
@matthew-andrews Can I merge this now, please?

commuterjoy:
This is nice, but my only advice is to think about idiot proofing this. It's really easy to set an env var, move directory, do a deploy in some other project and overwrite the VCL. Wondering if this might be an issue.

Perhaps specify the service ID as a parameter and leave the API key as an env var.

bin/next-build-tools.js
```diff
@@ -136,6 +137,19 @@ program
 		});
 
 	program
+		.command('deploy-vcl [folder]')
+		.description('Deploys VCL in [folder] to the fastly service given in the FASTLY_SERVICE_ID env var.  Also requires FASTLY_KEY env var')
+		.option('-m, --main <main', 'Set the name of the main vcl file (the entry point).  Defaults to "main.vcl"')
```
matthew-andrews: why not just specify the path to the VCL file directly?  and then default to wherever you think is best to put the main VCL file?

aintgoin2goa:
Good idea - have implemented it as an option to match the heroku cli

bin/next-build-tools.js
```diff
@@ -136,6 +137,20 @@ program
 		});
 
 	program
+		.command('deploy-vcl [folder]')
```
matthew-andrews: can we make this `deploy-vcl [file]` would be a bit simpler since this only support a single file anyway

aintgoin2goa: No - it supports a whole folder.  The code is taken from fastly-deploy which uploads a bunch of .vcl

matthew-andrews: ah ok

aintgoin2goa:
Can I merge now please?

matthew-andrews:
go for it

PR #77, opened 2015-04-27
added new task to verify all tricky dependency requirements

wheresrhys:
@matthew-andrews


matthew-andrews:
would you mind pulling this onto my v2 branch…… i've combined the verify stuff into a single command?

matthew-andrews:
might be worth changing `--skip-layout-checks` to `--skip-dependency-checks`

matthew-andrews:
will make v2 the canonical by the end of the day hopefully

PR #81, opened 2015-04-28
Build logging

ironsidevsquincy:
Log js/sass build completion

### DIFF
--- a/tasks/build.js
+++ b/tasks/build.js
@@ -15,6 +15,37 @@ var sourceFolder = './client/';
 var buildFolder = './public/';
 var mainJsSourceMapFile = 'main.js.map';
 
+var tasks = {
+	'build-sass': function() {
+		return obt.build.sass(gulp, {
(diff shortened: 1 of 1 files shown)


tasks/build.js
```diff
@@ -29,34 +60,18 @@ function run(task, opts) {
...
 		} else {
-			gulp.start.call(gulp, [task])
+			tasks[task]()
 				.on('end', resolve)
 				.on('error', reject);
 		}
 	});
 }
 
-gulp.task('build-sass', function() {
```
matthew-andrews: i quite liked that these were just gulp tasks when you got down to this level.  any particular reason to change this?

ironsidevsquincy: the tasks still exists, I've just stored the function in a variable so it can be shared above in `run` (`gulp.start` isn't really meant to be used, apparently, and non of the events were being fired anyway)

PR #87, opened 2015-04-29
Add install method so that projects don't need to directly depend on OBT anymore

matthew-andrews:


matthew-andrews:
this needs more thought……… somewhat circular.  we can't have nbt install rely on nbt being there :|

PR #91, opened 2015-04-30
Revert "Build logging"

ironsidevsquincy:
Reverts Financial-Times/next-build-tools#81


matthew-andrews:
what happened?

ironsidevsquincy:
fail when running `build-minify-js` due to it being a task rather than a function - fixing now

PR #92, opened 2015-04-30
Initial stub out of nbt run

matthew-andrews:


tasks/run.js
```diff
@@ -0,0 +1,49 @@
+'use strict';
+
+var exec = require('../lib/exec');
+var spawn = require('child_process').spawn;
+var packageJson = require(process.cwd() + '/package.json');
+var normalizeName = require('../lib/normalize-name');
+var LOCAL_PORT = 3002;
```
commuterjoy: Does this override the process.env.PORT and force every app to run on 3002?

matthew-andrews: at the moment yes

commuterjoy: ok, I wonder if that's going to be inconvenient. If production is set by `export PORT=1234` I think localhost should also be.

commuterjoy:
+1 looks good, apart from the port conversation

tasks/run.js
matthew-andrews: could make it overwriteable

commuterjoy: Just `var LOCAL_PORT = process.env.PORT || 3002;` ?

wheresrhys:
Nice. 

Can we look again at having a shared inherited makefile? When I was playing around with it a few months ago gulp was the stumbling block, not make

matthew-andrews:
@commuterjoy https://github.com/Financial-Times/next-build-tools/pull/92/files#diff-df650ebe12a0f6e1ca51c32e4158c2e5R53

matthew-andrews:
@wheresrhys you can if you like…  but unless it offers something drastically new and different I'm still in favour of keeping Makefile simple and moving the duplicated logic into nbt tasks.

PR #93, opened 2015-04-30
Correct callbacks

ironsidevsquincy:
And don't minify css if in dev mode

### DIFF
--- a/tasks/build.js
+++ b/tasks/build.js
@@ -2,6 +2,7 @@
 
 var gulp = require('gulp');
 require('gulp-watch');
+var yargs = require('yargs');
 
 var obt = require('origami-build-tools');
 var normalizeName = require('../lib/normalize-name');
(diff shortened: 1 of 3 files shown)


tasks/build.js
```diff
@@ -29,33 +30,39 @@ function run(task, opts) {
...
 gulp.task('build-sass', function() {
 	return obt.build.sass(gulp, {
-		sass: sourceFolder + mainScssFile,
-		buildFolder: buildFolder,
-		sourcemaps: true,
-		env: 'production'
-	});
+			sass: sourceFolder + mainScssFile,
+			buildFolder: buildFolder,
+			env: yargs.argv.dev ? 'development' : 'production'
```
matthew-andrews: hmm…… this should already been passed in from commander??

ironsidevsquincy: yeah, passed in https://github.com/Financial-Times/next-build-tools/blob/correct-dev-mode/tasks/build.js#L82, but args can't be passed to tasks. guess it could be made global at that point, but seemed messy. might be safer though

matthew-andrews: would prefer not to add another dependency …… :-/

PR #96, opened 2015-05-01
Fix purge command

aintgoin2goa:
remove proxy url and add a test

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,8 @@
+.PHONY: test
 SHOULD_BE = $(shell ./scripts/generate-docs.sh | md5 -q)
 IS = $(shell md5 -q README.md)
 
-test:
+verify:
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -10,5 +11,9 @@ else
 endif
 	./bin/next-build-tools.js verify --skip-layout-checks
 
+test:
+	mocha ./test/
```
matthew-andrews: not necessary to specify i think test is deafult

matthew-andrews:
i do believe this pr closes #71.  Probably means we need a new ticket “More tests!”, though

aintgoin2goa:
I think you might be pushing it to stay this completes closes that!  I could look at adding some more tests if you like (when I have the time).  Can I merge this?

matthew-andrews:
yep merge away.  don't worry about adding tests unless you need to fix a bit

PR #97, opened 2015-05-01
nbt scale

adgad:
Automatically scale a heroku app based on process information obtained from the service registry. 

/cc @matthew-andrews @richard-still-ft

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,6 +7,7 @@
 
     deploy [app]                                        runs haikro deployment scripts with sensible defaults for Next projects
     configure [options] [source] [target]               downloads environment variables from next-config-vars and uploads them to the current app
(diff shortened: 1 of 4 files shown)


matthew-andrews:
i like it

commuterjoy:
I was trying to think of a way to do this sort of thing with plain unix tools, but it's not that neat ...

`nbt list-services | xargs -I % heroku ps:scale web=% --app %` -- doesn't work :(

One thing we found useful Guardian is the concept of min and max application server pool sizes. So when something was under load we could type something like `gu up --apps article,fronts` to quickly give extra capacity for key applications. Having a concept of 'min' also prevented accidentally scaling down to 0.

matthew-andrews:
:+1: shipit

PR #100, opened 2015-05-06
Fix vcl vars

aintgoin2goa:
My code for inserting env vars into VCL prior to deploy wasn't working.  Have fixed it and added some tests


aintgoin2goa:
@matthew-andrews Any comments or can I go ahead and merge?

matthew-andrews:
nope looks good, thanks!

PR #104, opened 2015-05-11
Added deployTime to about

wheresrhys:
for diagnosing if your rebuild of a commit is being served or the previous deploy of that build

@matthew-andrews

### DIFF
--- a/lib/about.js
+++ b/lib/about.js
@@ -17,7 +17,8 @@ module.exports = function(opts) {
 				description: name,
 				support: 'next.team@ft.com',
 				supportStatus: 'active',
-				appVersion: commit
+				appVersion: commit,
+				deployTime: new Date().toString()
 			}));
 		});
 };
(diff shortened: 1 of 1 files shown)


matthew-andrews:
yup good idea :+1:

matthew-andrews:
strictly speaking though this isn't deploy time, it's build time.

matthew-andrews:
( in the future we might split those two things )

lib/about.js
```diff
@@ -17,7 +17,8 @@ module.exports = function(opts) {
 				description: name,
 				support: 'next.team@ft.com',
 				supportStatus: 'active',
-				appVersion: commit
+				appVersion: commit,
+				deployTime: new Date().toString()
```
commuterjoy: `new Date().toISOString()` is more parseable by future systems that might use this data

wheresrhys:
I was splitting that hair myself as soon as I'd committed. Will rename and ISO it up

matthew-andrews:
:+1:

PR #110, opened 2015-05-22
made about.json generation into a task

wheresrhys:


matthew-andrews:
cool — :+1: — out of curiosity what you using this for?

wheresrhys:
so s3 version of flags will have an about.json... I think/hope it'll work without an express instance

PR #113, opened 2015-06-05
Add --skip-gtg flag to deploy task

phamann:
Simply adds a flag `--skip-gtg` to the deploy task which in turn skips the good-to-go checks. This is for use cases where an app doesn't expose any HTTP endpoints, such as worker processes.

/cc @matthew-andrews @wheresrhys

### DIFF
--- a/tasks/deploy.js
+++ b/tasks/deploy.js
@@ -10,11 +10,11 @@ var logger = require('haikro/lib/logger');
 var normalizeName = require('../lib/normalize-name');
 var enablePreboot = require('../lib/enable-preboot');
 
-module.exports = function(app) {
+module.exports = function(opts) {
 	logger.setLevel('debug');
 	var token;
 	var commit;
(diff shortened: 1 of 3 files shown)


tasks/deploy.js
```diff
@@ -43,32 +43,36 @@ module.exports = function(app) {
...
-						});
-				}
-				checker = setInterval(checkGtg, 3000);
-				timeout = setTimeout(function() {
-					console.log("2 minutes passed, bailing");
-					reject(name + '.herokuapp.com/__gtg not responding with an ok response within 2 minutes');
-					clearInterval(checker);
-				}, 2*60*1000);
-			});
+			if(!opts.gtg) {
```
wheresrhys: rename this prop to `skipGtg` otherwise it looks like 'if not gtg do gtg'

phamann: Quite right :smile:

PR #116, opened 2015-06-11
Resolve deploy promise chain inside -skip-gtg

phamann:
/cc @wheresrhys @adgad

### DIFF
--- a/tasks/deploy.js
+++ b/tasks/deploy.js
@@ -73,6 +73,7 @@ module.exports = function(opts) {
 				});
 			} else {
 				console.log("Skipping gtg check");
+				return Promise.resolve();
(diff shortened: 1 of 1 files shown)


tasks/deploy.js
```diff
@@ -73,6 +73,7 @@ module.exports = function(opts) {
 				});
 			} else {
 				console.log("Skipping gtg check");
+				return Promise.resolve();
```
matthew-andrews: shouldn't be necessary?

phamann:
Opps thought I was on a different repo. Happy to revert. We are testing some weirdness we are seeing which could potentially be linked to this task.

PR #117, opened 2015-06-12
Revert "Resolve deploy promise chain inside -skip-gtg"

matthew-andrews:
Reverts Financial-Times/next-build-tools#116


matthew-andrews:
cc @phamann

PR #122, opened 2015-06-16
Skip enable preboot

matthew-andrews:


bin/next-build-tools.js
```diff
@@ -36,10 +36,12 @@ program
 	.command('deploy [app]')
 	.description('runs haikro deployment scripts with sensible defaults for Next projects')
 	.option('-s, --skip-gtg', 'skip the good-to-go HTTP check')
+	.option('-s, --skip-enable-preboot', 'skip the preboot')
```
ironsidevsquincy: both have the short form `-s`

matthew-andrews: bum

matthew-andrews: https://github.com/Financial-Times/next-build-tools/commit/ae73cef01354bdf18c5fe259a3a0986c3ec907cf

PR #124, opened 2015-06-18
Download single true source of environment variables from config vars provide them to run-local

matthew-andrews:


matthew-andrews:
Note this requires all developers have joined the `ft-next-config-vars` app on Heroku.

phamann:
@matthew-andrews How did I miss this?

PR #126, opened 2015-06-18
Fail the build when the build fails

matthew-andrews:
We came very very very close to breaking article today if it hadn't have been for a unit test that was checking `/public/main.css` and [image diff regressions](https://github.com/Financial-Times/next-article/pull/560) `nbt build` was not failing when the styling for an app couldn't be built.

See here:- https://travis-ci.org/Financial-Times/next-article/builds/67408615

The build errors on build of sass and merrily carries on as if nothing happened.

screen shot 2015-06-18 at 19 18 01

!!

### DIFF
--- a/tasks/build.js
+++ b/tasks/build.js
@@ -38,57 +38,45 @@ function run(task, opts) {
 }
 
 gulp.task('build-sass', function() {
-	return new Promise(function(resolve, reject) {
-		obt.build.sass(gulp, {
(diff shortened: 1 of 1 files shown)


matthew-andrews:
argh my implmentaiton totally broken here

matthew-andrews:
screen shot 2015-06-18 at 19 53 24


Success in the form of failure

PR #140, opened 2015-07-03
specify exact cache control for deploy static

wheresrhys:
because amazon erases previously set cache control if new asset uploaded without cache control set @matthew-andrews


matthew-andrews:
rubbery :+1: thsi might explain a few of @richard-still-ft issues

PR #144, opened 2015-07-09
Adding browsers to nightwatch tests

spriggsha:
@matthew-andrews 

SauceLabs is giving us 5 test runners instead of 3 now, and since we're spending a lot of time waiting for mobile simulators to spin up anyway, we can afford to add some more browsers without costing ourselves any additional time.

* Updating to latest chrome and firefox.
* Also adding latest-but-one for chrome and firefox, to put us inline with our official browser support policy.
* Adding ie9 and ie11.

### DIFF
--- a/config/nightwatch.json
+++ b/config/nightwatch.json
@@ -24,6 +24,20 @@
         "platform": "WINDOWS"
       }
     },
+    "chrome43": {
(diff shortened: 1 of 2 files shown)


matthew-andrews:
:ship: !

PR #149, opened 2015-07-17
scale to minimal number of dynos

wheresrhys:
So that make provision can have a go at starting workers in a branch (not sure if this would enable us to fail the build... might need some additional task to probe the logs)

@matthew-andrews @adgad


matthew-andrews:
normally in this i would just use the heroku cli or web ui.  is this something that we need to do regularly?

wheresrhys:
I'll come over and chat about the intention

PR #163, opened 2015-08-03
Fix capabilities for iPhone

ironsidevsquincy:
Also, name by device rather than OS, as sauce labs doesn't seem to quite think that way


matthew-andrews:
:+1:

PR #164, opened 2015-08-05
Router config

wheresrhys:
@matthew-andrews


wheresrhys:
Also see http://git.svc.ft.com/projects/NEXT/repos/router/pull-requests/50/diff

PR #175, opened 2015-08-11
more comprehensive url smoke tests

wheresrhys:
I took it out of gtg as it's useful to specify more config, and it blurs the meaning of gtg (as if it wasn't blurred enough!)

Example config here https://github.com/Financial-Times/next-ilyushin/blob/52e1240ae8a422e943d91c5ee8c0373fc294af80/test/smoke.js

### DIFF
--- a/bin/next-build-tools.js
+++ b/bin/next-build-tools.js
@@ -23,6 +23,7 @@ var downloadDevelopmentKeys = require('../tasks/download-development-keys');
 var run = require('../tasks/run');
 var about = require('../tasks/about');
 var rebuild = require('../tasks/rebuild');
+var testUrls = require('../tasks/test-urls');
 
 function list(val) {
 	return val.split(',');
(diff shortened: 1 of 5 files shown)


bin/next-build-tools.js
```diff
@@ -43,16 +44,32 @@ program
 	.option('--docker', 'deploy an app which uses docker')
 	.option('--gtg-urls <urls>', 'Comma separated list of urls to check before concluding the app is ok (these are in addition to __gtg)', list)
 	.action(function(app, options) {
+
+		if (options.gtgUrls && process.env.NODE_ENV !== 'production') {
```
matthew-andrews: node env?

matthew-andrews: nbt should never run in prod…… no need for this……

matthew-andrews:
one tiny comment otherwise :+1:

PR #186, opened 2015-08-20
Use Directly to throttle test-url requests

phamann:
Originally the `test-url` task was passing an array of fetches to `Promise.all` which in turn will fire all of the network requests at once. As the array could potentially be very large this was causing weird issues and requests to fail/timeout. 

This PR simply refactors the task to use @wheresrhys `Directly` promise throttle library and spread the requests out in batches of 10.

This seems to make the `next-article` tests pass: https://circleci.com/gh/Financial-Times/next-article/447

/cc @adambraimbridge @matthew-andrews

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,6 +11,7 @@
     "aws-sdk": "^2.1.19",
     "commander": "^2.6.0",
     "denodeify": "^1.2.0",
+    "directly": "^1.0.1",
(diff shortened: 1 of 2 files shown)


phamann:
The `next-article` build has consistently passed 3 times now with this change. So am happy to ship.

PR #194, opened 2015-08-25
get current keys from target, not source

aintgoin2goa:


matthew-andrews:
woah! how the hell has this worked for so long with this??

aintgoin2goa:
difference between the source and target keys is used to delete keys - and that would never be needed in a new branch app

matthew-andrews:
ah yes.  you're right.  thanks for figuring that out.

PR #195, opened 2015-08-25
test-urls stopped testing URLs

adgad:
/cc @ironsidevsquincy @matthew-andrews @aintgoin2goa 

(the diff looks worse than it is, I just removed the `return function()` line!)


matthew-andrews:
:grin:

matthew-andrews:
i hope this isn't the reason why we manaaged to fix article builds :see_no_evil:

PR #201, opened 2015-08-28
Document a sneaky workaround

matthew-andrews:


wheresrhys:
:+1: To avoid people relying on this hack longer than they should you could fail master builds if `.env` is present

matthew-andrews:
umm…… well once it's all set up properly nbt will overwrite the `.env` file :)

PR #205, opened 2015-09-02
Slow down test-urls

adgad:
/cc @wheresrhys @andygnewman @ironsidevsquincy 

Try and fix the failing article build


matthew-andrews:
:+1:

PR #209, opened 2015-09-02
Replace next express layout checks with n-layout ones

i-like-robots:
If an app uses n-layout this will now check if app's bower manifest contains duplicate deps.

Is this sensible @wheresrhys ?


wheresrhys:
:+1:

PR #210, opened 2015-09-04
Adding lastest FF and Chrome to nightwatch config

ironsidevsquincy:


adgad:
Shall we take out the older chrome/FFs? 

I believe our browser support is only latest - 1... @spriggsha ?

ironsidevsquincy:
Yeah, that's what I'm moving article over to, current & -1. Didn't want to remove any in case they're being used by projects (you can pass in the name of the browser when calling `nbt nightwatch`)

PR #218, opened 2015-09-21
allow vars to be loaded from .env file

aintgoin2goa:


bin/next-build-tools.js
```diff
@@ -207,6 +207,7 @@ program
 	.description('Deploys VCL in [folder] to the specified fastly service.  Requires FASTLY_KEY env var')
 	.option('-m, --main <main>', 'Set the name of the main vcl file (the entry point).  Defaults to "main.vcl"')
 	.option('-v, --vars <vars>', 'A way of injecting environment vars into the VCL.  So if you pass --vars AUTH_KEY,SERVICE the values {$AUTH_KEY} and ${SERVICE} in the vcl will be replaced with the values of the environmemnt variable')
+	.option('-e, --env', 'Load environment variables from local .env file (use when deploying from a local machine')
```
matthew-andrews: any reason why we don't want to just default to this?

aintgoin2goa: oops sorry - didn't see this comment. Then we would need to add --no-env or something for the CI deploy which seems odd to me.  I'm not really a fan of default=on as I think it's better to be explicit.  Then someone who looks at the makefile and sees and --env option can be clearer as to what is happening

PR #221, opened 2015-09-24
make smoke tests run consecutively

andygnewman:
On article smoke tests are frequently falling over with Heroku backlog too deep (H11), causing issues with build reliability. Suggest reducing the directly number to 1 to run the smoke tests consecutively. This will also aid debugging as will see which URL fails.


imranolas:
Didn't realise you'd done this already I've also created a PR #222. Probably better to for this value to be configurable.

andygnewman:
cc: @imranolas - agree - prefer your version, will close this.

PR #226, opened 2015-09-24
updated to obt 4.x.x and eslint

Lrnk:
@matthew-andrews whoops! Only realised you already had a branch for the beta obt4 when I tried to push this one!

I've gone for overriding just the bit of the obt eslint config we're temporarily unhappy with (the extreme var purge). I'm pretty sure that relative path to the obt version should work.


matthew-andrews:
if this works then :+1: i could never get the eslint stuff wroking :(

Lrnk:
refactored to obt4-stable2

PR #228, opened 2015-09-24
No more harmony

keirog:
Don't need it anymore


matthew-andrews:
:+1:

matthew-andrews:
good job we made a v4.0.0

matthew-andrews:
otherwise this would be an annoying breaking change

keirog:
exactly :smile:

PR #231, opened 2015-09-25
Add a disable nodemon flag

charypar:
For cases where you don't want server restarts on file change. One such case is front-end javascript hot loading. 

This shouldn't really affect anyone unless they opt in.

### DIFF
--- a/bin/next-build-tools.js
+++ b/bin/next-build-tools.js
@@ -228,6 +228,7 @@ program
 	.option('--procfile', 'Runs all processes specified in the Procfile')
 	.option('-s, --script <file>', 'Runs a single javascript file')
 	.option('--subargs [subargs]', 'Sub arguments to pass to a single script', /^\[.+]$/)
+	.option('--no-nodemon', 'Do not run through nodemon')
 	.action(function(opts){
 		run(opts).catch(exit);
 	});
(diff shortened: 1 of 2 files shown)


bin/next-build-tools.js
```diff
@@ -228,6 +228,7 @@ program
 	.option('--procfile', 'Runs all processes specified in the Procfile')
 	.option('-s, --script <file>', 'Runs a single javascript file')
 	.option('--subargs [subargs]', 'Sub arguments to pass to a single script', /^\[.+]$/)
+	.option('--no-nodemon', 'Do not run through nodemon')
```
matthew-andrews: hmm, i think we normally use `skip` as in `--skip-nodemon` for this… but sounds a little awkward so don't really mind if you keep as is

charypar: I also found it gets translated to a `nodemon` boolean option that is `true` by default. TJ thought of everything.

matthew-andrews: oh nice.

matthew-andrews:
:+1:

bin/next-build-tools.js
wheresrhys: for v4 of nbt I think we should use the `--no` prefix instead of every `--skip`. The conversion is so neat

PR #235, opened 2015-09-29
Fix source maps

ironsidevsquincy:
Previously we minified the JS within NBT, and pulled the source map out from an inline comment into an external file. This is now broken, e.g. the source map created looks like

```
{"version":3,"file":"main.js","sources":[],"names":[],"mappings":"","sourceRoot":"/source/","sourcesContent":[]}
```

However OBT4 has fixed the issue with minifying and creating a source map, so suggest we let OBT do its thing, inline the source map, and look at pulling it into another file in a future PR


ironsidevsquincy:
https://github.com/Financial-Times/origami-build-tools/issues/315

matthew-andrews:
won't this make our js files rather big?

ironsidevsquincy:
yes, but as we don't have working sourcemaps right now, better that than none.

keirog:
Another good use case for staging

matthew-andrews:
@keirog staging?

keirog:
Was a silly comment really, but I meant a staging environment with inline source maps (+ not included in prod)

wheresrhys:
OBT fix https://github.com/Financial-Times/origami-build-tools/pull/323

PR #242, opened 2015-09-30
allow modules on eslint

Lrnk:
This is allowed in OBT and but must not have been when I made the original copy. This inspired my request on origami's slack: https://financialtimes.slack.com/archives/ft-origami/p1443620468000382

### DIFF
--- a/bin/next-build-tools.js
+++ b/bin/next-build-tools.js
@@ -1,5 +1,4 @@
 #!/usr/bin/env node
-'use strict';
 
 require('isomorphic-fetch');
 
--- a/bin/next-build-tools.js
+++ b/bin/next-build-tools.js
@@ -1,4 +1,5 @@
 #!/usr/bin/env node
+'use strict';
 
 require('isomorphic-fetch');
 
--- a/config/eslint.json
+++ b/config/eslint.json
@@ -6,7 +6,7 @@
 	"node": true
   },
   "ecmaFeatures": {
-	"modules": false
+	"modules": true
   },
   "rules": {
 	"no-unused-vars": 2,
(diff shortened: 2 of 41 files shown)


matthew-andrews:
ok

Lrnk:
@matthew-andrews are you also cool with the removal of use-strict throughout? eslint demands it, as apparently strict mode is implied for modules (es6 modules and node modules) (my slightly buried source: https://github.com/google/traceur-compiler/issues/1031)

ironsidevsquincy:
Found this module feature to be a real pain. This change will make us remove `use strict`, but we're not using modules in our server-side node code, still using good old `require`s (CommonJS), which don't imply `use strict`, so we do need to have them.

Lrnk:
@ironsidevsquincy ahh I see I didn't realise. Without it linting fails on any projects we do use es6 modules. I suppose we'll have to convert to es6 modules for all our projects at once or not at all :anguished:

matthew-andrews:
is there a laissez faire option?

Lrnk:
I had a go at laissez faire by leaving this as it is and overriding the es6 modules rule per-file but apparently that doesn't work for es6 stuff

ironsidevsquincy:
@Lrnk maybe two verifies, one for `client`, one for `server`, set the modules for the former, and also get everyone to switch over to `imports`

Lrnk:
That sounds like a sensible solution.

I did just think of a faster but less sensible solution though - we could relax the strict mode rule so it doesn't mind us leaving use strict in after all

ironsidevsquincy:
yeah, that one. until everyone's on the same page engine-wise, it's going to be messy so might as well do the easiest option for now

Lrnk:
Cool, cheers dudes. Happy for me to merge?

Lrnk:
Assuming yes

PR #253, opened 2015-10-02
Allow passing of custom script in conjunction with run --local

phamann:
As I have the need to do this. 

/cc @matthew-andrews


matthew-andrews:
:+1:

PR #257, opened 2015-10-06
Better float

aintgoin2goa:
By default the float task destroyed the test app after running gtg checks - this is now optional so smoke tests can be run.  Also pulled in latest changes to master and fixed an issue with babel/obt/es7.

### DIFF
--- a/tasks/float.js
+++ b/tasks/float.js
@@ -10,6 +10,7 @@ const log = require('../lib/logger');
 
 
 module.exports = function(opts){
+	console.log('float', opts);
 	var testAppName;
 	return co(function* (){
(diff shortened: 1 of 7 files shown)


tasks/float.js
```diff
@@ -39,16 +40,23 @@ module.exports = function(opts){
...
 		log.error('Man overboard', err);
-		return destroy({app:testAppName, verbose:true}).then(function(){
-			throw err;
-		});
+		if(opts.destroy){
+			return destroy({app:testAppName, verbose:true}).then(function(){
+				throw err;
+			});
+		}else{
+			return Promise.resolve(null);
```
matthew-andrews: not necessary :-)

PR #258, opened 2015-10-07
WIP Make jira-cmd a cli tool

commuterjoy:
- [ ] Add some canned commands to create and list next bugs

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,7 +3,8 @@
   "version": "4.3.0",
   "bin": {
     "next-build-tools": "./bin/next-build-tools.js",
-    "nbt": "./bin/next-build-tools.js"
+    "nbt": "./bin/next-build-tools.js",
+    "jira": "./node_modules/jira-cmd/bin/jira.js"
   },
   "scripts": {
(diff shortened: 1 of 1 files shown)


matthew-andrews:
does this need to be part of NBT?

package.json
```diff
@@ -3,7 +3,8 @@
   "version": "4.3.0",
   "bin": {
     "next-build-tools": "./bin/next-build-tools.js",
-    "nbt": "./bin/next-build-tools.js"
+    "nbt": "./bin/next-build-tools.js",
+    "jira": "./node_modules/jira-cmd/bin/jira.js"
```
matthew-andrews: npm does not guaranteed this path will always work……

commuterjoy: ok, good tip

commuterjoy:
> does this need to be part of NBT?

Not especially :)

I could just document it on the new starter guide as an optional install.

matthew-andrews:
> optional install
let's do that … NBT slows down builds already so much … trying to pare it down to the minimum ……

PR #263, opened 2015-10-08
add skip-logging option to ship

aintgoin2goa:


bin/next-build-tools.js
```diff
@@ -306,7 +306,9 @@ program
 	.option('-s --no-scale', 'Skip the scale step')
 	.option('-p --pipeline [name]', 'The name of the pipeline to deploy to.  Defaults to the app name')
 	.option('-m --multiregion', 'Will expect a US app as well as an EU one')
+	.option('-l --skip-logging', "Don't log to Salesforce™®©")
```
matthew-andrews: ahh did you want this to be `no-logging` ?

matthew-andrews:
this all looks good to me, added a comment — think you suggested (and i liked the suggestion) to eventually move `--skip-` to `--no-`

aintgoin2goa:
I don't think it was me, but it's a good suggestion

PR #270, opened 2015-10-14
Pick a standard for function () vs. function()

matthew-andrews:
@wheresrhys

### DIFF
--- a/config/eslint.json
+++ b/config/eslint.json
@@ -25,6 +25,7 @@
 	"one-var": [2, "never"],
 
 	"no-var": 1,
+	"space-before-function-paren": [1, "always"],
 	"strict": [1, "global"]
   },
   "globals": {


ironsidevsquincy:
Vote for amendment:

{"anonymous": "always", "named": "never"}

cf: http://javascript.crockford.com/code.html#function

keirog:
:+1:

matthew-andrews:
wait so:-

```js
module.exports = function() {
  myfunc();
};

function myfunc() {
  alert("yup i'm using an alert");
}
```

is bad?

and this is good?

```js
var myfunc = function() {
  alert("yup i'm using an alert");
};

module.exports = function() {
  myfunc();
};
```

ironsidevsquincy:
### Bad
```
function() {
    ...
}

function foo () {
   ...
}
```

### Good
```
function () {
    ...
}

function foo() {
   ...
}
```

matthew-andrews:
What's wrong with:-

```js
function foo() {
}

function() {
}
```

matthew-andrews:
My eyes see the inconsistency next the brackets (there should also be text up against the `(`) … and `function() { }` + `function foo() { }` feels the tightest… least wasteful extra characters…

ironsidevsquincy:
To quote DC, "If a function literal is anonymous, there should be one space between the word function and the ( (left parenthesis). If the space is omitted, then it can appear that the function's name is function, which is an incorrect reading."

Which seems a valid reason to do it.

PR #272, opened 2015-10-15
App's tests are in a 'test' dir

ironsidevsquincy:


matthew-andrews:
:+1: 
i think they used to be in `tests`

PR #289, opened 2015-10-22
adds ability to run eslint fix via nbt verify

wheresrhys:
To make it easier to upgrade repos from es5 to es6 `nbt verify --fix` can be called to auto fix warnings and errors


wheresrhys:
actually --fix seems pretty impotent when it comes up against almost all linting errors

PR #290, opened 2015-10-22
add nodemon to deps list

aintgoin2goa:
`nbt run --watch` uses this but it's not actually required..

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,6 +24,7 @@
     "haikro": "^2.2.0",
     "isomorphic-fetch": "^2.0.0",
     "next-gulp-tasks": "Financial-Times/next-gulp-tasks#v1",
+    "nodemon": "^1.7.3",
     "origami-build-tools": "^4.2.0",
     "semver": "^5.0.3",
     "shellpromise": "^1.0.0"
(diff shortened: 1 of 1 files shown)


matthew-andrews:
:+1:

matthew-andrews:
oops :innocent:

matthew-andrews:
actually this might not work

matthew-andrews:
i don't think the `bin` scripts of dependencies get created…

matthew-andrews:
i think we'd need to use this:-
https://github.com/remy/nodemon/blob/master/doc/requireable.md

and basically copy the contents of this into NBT
https://github.com/remy/nodemon/blob/master/bin/nodemon.js

and add `nodemon` to `package.json` scripts https://github.com/Financial-Times/next-build-tools/blob/master/package.json#L4-L7

aintgoin2goa:
I'm confused.  Surely the first link tells us that we don't need the other 2 things??

We would just require nodemon and run it there and then...

The other alternative is to require it to be installed globally...

matthew-andrews:
The first option might work… worth trying… we might need to do something to isolate the two `nodemon`s that will be run (the local app and the router app) from each other though.

aintgoin2goa:
I don't know what this is anymore, but it's probably stale.  Closing...

PR #295, opened 2015-10-28
using curl to test redirects in order to get no-follow

wheresrhys:


matthew-andrews:
wait

matthew-andrews:
probably obvious but why can't use `follow: 0` here?
https://github.com/bitinn/node-fetch#options

wheresrhys:
Had this chat before. node-fetch/fetch/request all have the behaviour that they're fine up until you hit the redirect limit, but then error as soon as you hit it, so you can't read a 302 response without following it

matthew-andrews:
ah yes sorry

matthew-andrews:
:+1:

PR #298, opened 2015-10-30
Make console.log a linting warning

charypar:
Because people should use logger.

@matthew-andrews


matthew-andrews:
dat woz qwik

matthew-andrews:
:+1:

charypar:
I don't have much time left :)

PR #302, opened 2015-11-03
fixed post smoke requests

laurieboyes:
`-I` and `-d` params are incompatible (attempts to send a request for the headers and a separate one for the post)

`-D-` has the same effect as `-I` as far as nbt is concerned

(http://stackoverflow.com/a/287299)

### DIFF
--- a/tasks/test-urls.js
+++ b/tasks/test-urls.js
@@ -139,14 +139,14 @@ class UrlTest {
 
 
 	getRedirect () {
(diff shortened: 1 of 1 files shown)


adgad:
Wouldn't this affect GETs as well?

laurieboyes:
They still work with the `-D-` thing and I reckoned it was better to keep them consistent

laurieboyes:
Got a cautious IRL :+1: from @adgad

PR #306, opened 2015-11-06
Set outbound cache metadata

ironsidevsquincy:
Which will produce a header of the format `x-amz-meta-outbound-cache-control: ...`

See http://git.svc.ft.com/projects/NEXT/repos/fastly-config/pull-requests/100/overview


ironsidevsquincy:
no needed as assets back on geebee

PR #317, opened 2015-12-10
fix git tag sort

laurieboyes:
`semver.gt` returns a boolean. swap for `semver.compare` which returns -1, 0 or 1, and will work as a comparator function
@wheresrhys


wheresrhys:
:+1::+1::+1::+1::+1::+1::+1::+1::+1::+1::+1::+1::+1::+1::+1::+1:

PR #319, opened 2015-12-18
Be able to override the entry path to Sass/Js compilation

matthew-andrews:
Enables @commuterjoy to easily compile separate JS/CSS files.  Includes the delightful new option `--skip-about`.

To build a file located in `client/mattc.js` in the `build-production` step run:-

```sh
$ nbt build --main-js mattc.js --skip-sass --skip-about --skip-haikro --skip-hash
```

And in the `make build` step it's the same with `--dev` added:-

```sh
$ nbt build --main-js mattc.js --skip-sass --skip-about --skip-haikro --skip-hash --dev
```

I think in the next major version of NBT I should switch the logic to be opt-in rather than opt-out :grin:.

### DIFF
--- a/tasks/build.js
+++ b/tasks/build.js
@@ -97,7 +97,7 @@ gulp.task('build-worker', function() {
 });
 
 
-function task (opts) {
+function task(opts) {
 	isDev = opts.isDev;
 	let tasks = [];
 	if (!opts.skipSass) {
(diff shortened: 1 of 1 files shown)


matthew-andrews:
```
5.14.0 published to npm and tagged in git
```

matthew-andrews:
Note you should run this command BEFORE running the main `nbt build`

PR #322, opened 2016-01-07
Fix vcl deploying to handle fastly API change

adgad:
*undocumented/unversioned API change https://docs.fastly.com/api/config#service

/cc @matthew-andrews @commuterjoy


matthew-andrews:
how rude of them :+1:

PR #324, opened 2016-01-08
bottle handles prereleases

wheresrhys:
@tom-parker

### DIFF
--- a/tasks/bottle.js
+++ b/tasks/bottle.js
@@ -1,13 +1,15 @@
-
 'use strict';
 
 const shell = require('shellpromise');
 const semver = require('semver');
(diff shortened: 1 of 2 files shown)


tasks/bottle.js
```diff
@@ -1,13 +1,17 @@
-
 'use strict';
 
+console.log('THIS IS THE CORRECT NBT')
```
tom-parker: This intentional?!

tom-parker:
:100:

tasks/bottle.js
wheresrhys: haha. no

PR #328, opened 2016-01-18
Added some post-bottle advice (console log)

adambraimbridge:
cc @matthew-andrews

### DIFF
--- a/tasks/bottle.js
+++ b/tasks/bottle.js
@@ -191,6 +191,7 @@ function bottle(versions, increment, forceNpm) {
 	const tagVersion = versions[1];
 	const bowerVersion = versions[2];
 	const packageVersion = versions[3];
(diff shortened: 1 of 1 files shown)


matthew-andrews:
weird error :+1:

tasks/bottle.js
```diff
@@ -191,6 +191,7 @@ function bottle(versions, increment, forceNpm) {
 	const tagVersion = versions[1];
 	const bowerVersion = versions[2];
 	const packageVersion = versions[3];
+	const postBottleAdvice = `\nNote: To update this code in the app/s that use it, use 'nbt rebuild {app name}', or 'nbt rebuild --serves {service type}'`;
```
keirog: There's a scope issue here, hence the failing build. Declare `let postBottleAdvice;` outside of the bottle function (i.e. in the main body of the module), then just set the val on line 194 without `const`. That should fix it.

PR #331, opened 2016-01-19
Add no-const-assign to eslint config

### LINKED ISSUE #330: Propose adding "no-const-assign" as an eslint rule
Reassigning a const is a runtime error in ES6 but `nbt verify` doesn't even warn about it.

Can we add [no-const-assign](http://eslint.org/docs/rules/no-const-assign) as a hard error (level 2)?

callumlocke:
fixes #330

### DIFF
--- a/config/eslint.json
+++ b/config/eslint.json
@@ -13,6 +13,7 @@
     "no-unused-vars": 2,
     "no-undef": 2,
     "eqeqeq": 2,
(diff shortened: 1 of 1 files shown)


matthew-andrews:
i'm cool with this.  can you make give people a bit of a warning in #ft-next-dev and get a few other :+1:s?

PR #337, opened 2016-02-12
allow running as https

lc512k:
Allow alternative rule for running as https by passing subargs options to the proxy, i.e.
```
run:
	export DEBUG=ft-next-signup-debug; \
	nbt run --harmony
run-https: 
	export DEBUG=ft-next-signup-debug; \
	nbt run --harmony --subargs [--https,key.pem,key-cert.pem]
```

cc/ @matthew-andrews @wheresrhys 

Relevant router update in:
http://git.svc.ft.com/projects/NEXT/repos/router/pull-requests/67/diff#server/proxy.js

cc/ @ifyio This is what we need to run https://local.ft.com/signup to integrate with the Zuora iframe

### DIFF
--- a/tasks/run.js
+++ b/tasks/run.js
@@ -101,32 +101,17 @@ function runRouter(opts) {
 		PORT: opts.PORT
 	};
 
-	var https = false;
-
-	// for HTTPS we expect subargs as [--https,path/to/cert.pem,path/to/key.pem]
-	if (opts.subargs) {
-		
-		var args = opts.subargs.replace(/^\[/, '').replace(/]$/, '').split(',');
-
-		var https = args.length && args[0].indexOf('https') !== -1 ? args[0] : false;
-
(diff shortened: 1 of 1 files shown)


wheresrhys:
Is there some reason for not doing `nbt run --harmony --https [key.pem,key-cert.pem]`? 

otherwise looks good

wheresrhys:
Just getting my head round both PRs.

So in the other one (running the router) you need to pass all these settings in as subargs because you're using `nbt run --script`.

But am I right in thinking this PR is for starting any pp on https so you can run it through the https router? If so, would `nbt run --harmony --https --key key.pem --cert key-cert.pem` be clearer?

lc512k:
@wheresrhys, yes, `nbt run --harmony --https --key key.pem --cert key-cert.pem` would do just fine. I'll update :+1:

wheresrhys:
:+1:

PR #338, opened 2016-02-17
add default tag version to bower bottle

andygnewman:
When doing a bottle v1 for a bower component it tagged it as `vnull` and failed on the ASCII art.

This change should default to v1.0.0 if no existing release version found.

cc: @wheresrhys

### DIFF
--- a/tasks/bottle.js
+++ b/tasks/bottle.js
@@ -178,6 +178,8 @@ function getFixedVersion (currentVersion, proposedVersion) {
 				return /^v/.test(proposedVersion) ? proposedVersion : ('v' + proposedVersion);
 
 			}
+		} else if (proposedVersion === 'v1') {
+			return 'v1.0.0';
 		}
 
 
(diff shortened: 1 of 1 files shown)


wheresrhys:
have you tested with bower components, npm components, npm + bower components? I'm not surprised it's broken, as it's complex, but it's important to be careful about not accidentally releasing to npm when not needed

wheresrhys:
actually having read your fix, I think the problem is that fixedVersion should be defined as `v1.0.0` by this point, so the problem lies around about here https://github.com/Financial-Times/next-build-tools/blob/master/tasks/bottle.js#L147 let me know if you want to talk it through

andygnewman:
Yep, makes sense to fix it there. I'll take a look in the morning and see if I can propose something a bit more robust.

wheresrhys:
also, to unblock whatever else you were working on, for bower components you can just release a tag in github (although in general I'd like to discourage this as sometimes components are both bower & npm but people don't realise)

wheresrhys:
:+1:

PR #346, opened 2016-03-15
Stop redness

aintgoin2goa:
Whenever nbt errors it makes everything else in that console window be red.  Got rid of the hack and replaced it with colors


matthew-andrews:
+1

PR #348, opened 2016-03-23
Remove preboot enabling from apps

matthew-andrews:
To fix build breakages… hopefully…


ironsidevsquincy:
:+1:

keirog:
:+1:

matthew-andrews:
i'm going to try this out on front page if that's ok?

PR #349, opened 2016-03-23
Revert "Remove preboot enabling from apps"

matthew-andrews:
Reverts Financial-Times/next-build-tools#348


matthew-andrews:
urgh wrong one

PR #353, opened 2016-04-07
Update download-development-keys.js

adambraimbridge:
`config-vars` moved from bitbucket to github (private) so changed URL accordingly


matthew-andrews:
+1

PR #357, opened 2016-04-19
pass retries and suiteRetries options through if present

lc512k:
Pass `retries` and `suiteRetries` options through if present
cc @matthew-andrews


matthew-andrews:
+1

matthew-andrews:
fyi there's been a major version bump

lc512k:
Merging so it goes out with the unit test fix

PR #358, opened 2016-04-19
remove @financial-times from appName

lc512k:
@matthew-andrews this fixes the broken test, but I'm entirely not sure if this could have unwanted repercussions elsewhere?


matthew-andrews:
hmm yeah let me look at this

PR #371, opened 2016-05-09
Smoke test are now required for apps

wheresrhys:
@matthew-andrews @TheoLeanse 

Also @aintgoin2goa Can you think of anywhere else I'd need to add this for ship/flat etc. They all look like they only ever use deploy internally, but not suer if I missed something


tasks/deploy.js
```diff
@@ -55,11 +65,7 @@ module.exports = function (program, utils) {
 		.command('deploy [app]')
 		.description('runs haikro deployment scripts with sensible defaults for Next projects')
 		.option('-s, --skip-gtg', 'skip the good-to-go HTTP check')
-		.option('--gtg-urls <urls>', 'Comma separated list of urls to check before concluding the app is ok (these are in addition to __gtg)', utils.list)
```
matthew-andrews: could we have a `--skip-smoke` option for cases so we can opt out of this feature?

wheresrhys: why? which cases?

matthew-andrews: apps that don't have web processes, e.g. flag notifier

wheresrhys: in that case --skip-gtg covers it

wheresrhys: which I could alias to --skip-smoke-tests

matthew-andrews: ah ok yep

matthew-andrews: well

matthew-andrews: yes i guess so

PR #373, opened 2016-05-09
add backend authentication to smoke tests against staging

wheresrhys:
dependent on this https://github.com/Financial-Times/next-config-vars/pull/128/files

@ironsidevsquincy @aintgoin2goa

### DIFF
--- a/lib/smoke-test.js
+++ b/lib/smoke-test.js
@@ -7,6 +7,7 @@ const packageJson = require(process.cwd() + '/package.json');
 const shell = require('shellpromise');
 
 let baseUrl;
+let authenticate = false;
 
 function getExpectation (expected) {
 	if (typeof expected === 'string') {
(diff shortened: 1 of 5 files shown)


lib/smoke-test.js
```diff
@@ -189,14 +193,16 @@ function testUrls (opts) {
...
-
 function runTest (opts) {
 	const appName = (opts.app) ? opts.app : 'ft-next-' + normalizeName(packageJson.name);
 	const urlConfig = require(path.join(process.cwd(), opts.configPath || 'test/smoke.js'));
 	baseUrl = 'http://' + appName;
 	if (!/:|\./.test(appName)) {
 		baseUrl += '.herokuapp.com';
 	}
+	if (opts.authenticated) {
+		authenticate = true;
```
ironsidevsquincy: rather than globally setting, doesn't the `UrlTest` constructor have access to this `opts`, and therefore the `authenticated` flag?

wheresrhys: no that's a different opts. This does feel ugly, but it'd need quite a refactor to pass it around properly.

PR #380, opened 2016-05-16
don't allow smoke test cheating

wheresrhys:
@matthew-andrews


matthew-andrews:
you're a mean monopoly player.  i bet you have hotels on mayfair and parklane.

PR #383, opened 2016-05-19
Skip gtg check during prod deploys

matthew-andrews:
@wheresrhys


wheresrhys:
cool. :+1: if we're sure this won't end up applying to staging apps

matthew-andrews:
yah this only gets run when calling `deploy` CLI command directly, the  task required in ship is in the layer behind this

PR #384, opened 2016-05-20
read saucelabs browser list from env var

wheresrhys:
@matthew-andrews @bjfletcher


matthew-andrews:
o.O  this would be set in continuous-integration field of config vars?

wheresrhys:
yip

PR #387, opened 2016-05-23
filtering by unstable or unstable js browser

wheresrhys:
@lc512k


tasks/nightwatch.js
```diff
@@ -12,13 +12,31 @@ function toStdErr(data) {
 	process.stderr.write(data.toString());
 }
 
+function filterBrowsers (env, broken) {
+	if (!broken) {
```
lc512k: can we log a warning with the browsers we're filtering out? So devs know their test didn't actually run on all the browsers they expected it to

wheresrhys: yep good idea

Issue #388, opened 2016-05-23
deploy-static exits silently when given a directory source


lucas42:
If the source parameter of deploy-static is a single directory, nothing happens.

**Steps to reproduce:**
```
mkdir build
nht deploy-static build
```
(Also occurs if you put some stuff in the directory)
**Output:**
No text.  Exit code 0.

**Expected:**
Some indication of what just happened.

matthew-andrews:
PRs welcome 🙈

PR #390, opened 2016-05-23
Add warning to deploy-static if no files are found

lucas42:
Addresses #388

### DIFF
--- a/tasks/deploy-static.js
+++ b/tasks/deploy-static.js
@@ -18,6 +18,10 @@ function task (opts) {
 	const destination = opts.destination || "";
 	const bucket = opts.bucket;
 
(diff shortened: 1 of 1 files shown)


tasks/deploy-static.js
```diff
@@ -18,6 +18,10 @@ function task (opts) {
 	const destination = opts.destination || "";
 	const bucket = opts.bucket;
 
+	if (files.length < 1) {
+		console.warn("Warning: No files found for upload to s3.  (Directories are ignored)");
```
matthew-andrews: `return Promise.reject('No files found for upload to s3.  (Directories are ignored)');`

matthew-andrews:
+1

PR #393, opened 2016-05-25
Add ability to monitor assets when deploying

ironsidevsquincy:
With `--monitor-assets`, sends asset size/gzip size to Graphite

Dependent on https://github.com/Financial-Times/next-metrics/pull/116


tasks/deploy-hashed-assets.js
```diff
@@ -37,7 +38,8 @@ function upload(params) {
 			});
 }
 
-function task (app) {
+function task (opts) {
```
ironsidevsquincy: think this is ok, couldn't find any apps that were calling this directly with an app name

PR #402, opened 2016-07-07
apply surrogate-control to static assets

wheresrhys:


wheresrhys:
@matthew-andrews @phamann because when we move geebee to fastly we can have better control of headers

matthew-andrews:
that's a good idea

PR #403, opened 2016-07-14
Add the ability to set multiple local apps

ironsidevsquincy:
Not just the one currently running, e.g.

`nht run --local-apps service-worker=3010,front-page=3011`


matthew-andrews:
ha very cool!

PR #406, opened 2016-07-25
add reporter option

lc512k:
cc @TheoLeanse  @matthew-andrews

Add `--reporter` option to nightwatch for custom reporting


TheoLeanse:
Looks good to me!

PR #408, opened 2016-07-27
Ensure .env file is loaded for smoke tests

laurieboyes:
@wheresrhys 

Naive (but possible acceptable?) approach to ensuring the .env file is loaded for smoke tests, so that we can use env for to pass authentication for routes we want to test

### DIFF
--- a/lib/smoke-test.js
+++ b/lib/smoke-test.js
@@ -200,6 +200,7 @@ function testUrls (opts) {
 }
 
 function runTest (opts) {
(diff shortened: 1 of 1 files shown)


keirog:
What's the risk here of outputting our keys to circle logs?

wheresrhys:
isn't it already loaded prior to running any nht task?

laurieboyes:
This was never necessary. There was a gap in my CI knowledge that has now been filled.

PR #414, opened 2016-08-15
set encoding for JS and CSS assets to be unicode

bjfletcher:
@wheresrhys @i-like-robots @ironsidevsquincy hopefully this will be it


wheresrhys:
This will also need adding to the deploy-static task I think (which is how __sw.js and n-ui.js get deployed)

bjfletcher:
@wheresrhys I've added that
and also I've asked it to read files using unicode

wheresrhys:
:+1::+1:

PR #447, opened 2017-02-27
don't watch public folder in nodemon

wheresrhys:


tasks/run.js
```diff
@@ -78,6 +78,7 @@ function runLocal(opts) {
 		}
 
 		if(opts.nodemon) {
+			args.push('--ignore', 'public/');
```
leggsimon: Most apps' `nodemon.json` includes `public/head.css` do we want to include that or just ignore `/public` completely? It seemed to only be `public/about.json` that caused all the restarts

ironsidevsquincy: I don't think we need to manually listen to `head.*` changes anymore, as n-express was updated to monitor changes. Though I've no idea where that is now...

PR #448, opened 2017-04-25
add --no-configure option for float like with ship

bjfletcher:


bjfletcher:
@wheresrhys looks like eslint has been switched on for this project but the errors haven't been fixed, quite a lot of them. I'll leave this for some other day.

PR #451, opened 2017-05-11
Check routes

lc512k:
@wheresrhys this check is brilliant (checking smoke.js includes all of that app's routes if it's user-facing) but the conditional was being skipped:

now it checks that:

`http://ft-next-signup-eu.herokuapp.com` contains `ft-next-signup`, rather than the other way around

Also, I have a feeling this will blow up all the things. Which isn't terrible. We need this.

### DIFF
--- a/lib/smoke-test.js
+++ b/lib/smoke-test.js
@@ -35,6 +35,8 @@ function getExpectation (expected) {
 
 class UrlTest {
 	constructor (opts) {
+
+
 		if (opts.url === '/__gtg') {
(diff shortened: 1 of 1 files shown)


wheresrhys:
haha thanks. I don't think it's ever worked though. Maybe your change will fix it :+1:

lc512k:
@wheresrhys I've made it a warning for now to give people a chance to add their routes into smoke.js. Will add `throw new Error` back in soon.

PR #453, opened 2017-05-24
Normalize Vault names

sjparkinson:
There's going to be a mix of app names, `spoor-foobar`, `ft-next-foobar`, `foobar`, `next-foobar`

Normalize the name, then get the path from the registry.

### DIFF
--- a/tasks/configure.js
+++ b/tasks/configure.js
@@ -26,9 +26,16 @@ function fetchFromNextConfigVars(source, target, key) {
 
 function fetchFromVault(source, target) {
 	console.log(`Fetching ${source} config from the vault for ${target}`);
-	const pathPrefix = 'secret/teams/next';
-	return vault.get()
-		.then(vault => vault.read(`${pathPrefix}/${source}/production`))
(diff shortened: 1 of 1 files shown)


tasks/configure.js
```diff
@@ -26,9 +26,16 @@ function fetchFromNextConfigVars(source, target, key) {
...
 	console.log(`Fetching ${source} config from the vault for ${target}`);
-	const pathPrefix = 'secret/teams/next';
-	return vault.get()
-		.then(vault => vault.read(`${pathPrefix}/${source}/production`))
+
+	const normalizedSource = normalizeName(source);
+
+	const path = fetch('https://next-registry.ft.com/v2/')
+		.then(fetchres)
+		.then(json => json.find(app => app.name === normalizedSource).config)
```
bjfletcher: I'd get rid of `const normalizedSource = normalizeName(source);` line and modify this line to become:

`.then(json => json.find(app => app.name === normalizeName(source)).config)`

PR #455, opened 2017-05-26
allow commands to specify a custom `--registry`

constantology:


tasks/configure.js
```diff
@@ -24,10 +26,10 @@ function fetchFromNextConfigVars(source, target, key) {
 		});
 }
 
-function fetchFromVault(source, target) {
+function fetchFromVault(source, target, registry = DEFAULT_REGISTRY_URI) {
```
bjfletcher: let's ask in #ft-next-dev if we can stop supporting Node 4 with NHT

tasks/configure.js
```diff
@@ -126,6 +128,7 @@ module.exports = function (program, utils) {
 		.description('gets environment variables from next-config-vars or the vault and uploads them to the current app')
 		.option('-o, --overrides <abc>', 'override these values', utils.list)
 		.option('-n, --no-splunk', 'configure not to drain logs to splunk')
+		.option('-r, --registry [registry-uri]', `use this registry, instead of the default: ${DEFAULT_REGISTRY_URI}`, DEFAULT_REGISTRY_URI)
```
bjfletcher: nice :)

PR #456, opened 2017-05-30
Add option to retrieve config-vars from non-production source

alexnaish:


tasks/configure.js
```diff
@@ -10,8 +10,15 @@ var fetchres = require('fetchres');
 
 function fetchFromNextConfigVars(source, target, key) {
 	console.log(`Fetching ${source} config from Next Config Vars for ${target}`);
-	return fetch('https://ft-next-config-vars.herokuapp.com/production/' + source, { headers: { Authorization: key } })
+	const configEnv = process.env.CONFIG_ENV || 'production';
```
sjparkinson: Thoughts on making this into a CLI option? Would make it a self-documenting feature then, and you might be able to validate that the value is valid too.

alexnaish: Happy to do that, the only issue I can see is that we don't call configure directly - we call it via float / ship which would mean we would need to pass options down through those tasks that are only related to the configure stage.

PR #460, opened 2017-06-01
Remove references to config-env

alexnaish:


andygout:
Have prepared a PR for removing that switch from the Makefile here: https://github.com/Financial-Times/next-retention/pull/83

PR #461, opened 2017-06-07
assume pipelines supported

adgad:


aintgoin2goa:
Pipelines is still a CLI plugin I think - presumably there is a bit in the CI setup that installs this, but might be worth bearing in mind in case that ever gets removed...

PR #463, opened 2017-06-09
Allow a component to deploy hashed assets

wheresrhys:
Which means (with a bit of additional work) n-ui shoudl be servable on hashed urls


leggsimon:
Would you mind just commentting what the regexes are checking for?

leggsimon:
Thanks for that 😄

PR #470, opened 2017-07-06
Run router only

aintgoin2goa:
So this is actually a small change.  But I had to fix tons of linting errors before I was allowed to commit.

All I've done is add the option `--no-app` to the `nht run`.  This makes is start the router only.

This is so I can run the app myself in a debugging session and set breakpoints and stuff.


aintgoin2goa:
Found better ways of doing this, so not really useful

PR #471, opened 2017-07-07
If we can't deploy - fail

aintgoin2goa:


aintgoin2goa:
Our deploys were failing but the script was returning exit code 0 - this should fix that

PR #474, opened 2017-08-02
Logging s3 upload errors

adgad:


tasks/deploy-hashed-assets.js
```diff
@@ -31,10 +31,12 @@ function task (opts) {
 
 	function upload(params) {
 		return new Promise((resolve, reject) => {
-			return s3bucket.upload(params, err => {
+			return s3bucket.upload(params, (err, data) => {
 						if (err) {
+							console.error('Upload failed', err);
 							reject(err);
 						} else {
+							console.log('Upload success', data);
```
wheresrhys: is there a need to log data?

adgad: My assumption is that, given it's getting to the polling of URLs, that it's not actually erroring and this promise is resolving.

PR #475, opened 2017-08-24
add router type option

leggsimon:
🐿 v2.5.16

This adds an option to specify which local router to use. use case for this is to be able to use `kat-router` like `next-router`, nice and easily.

Related: https://github.com/Financial-Times/kat-router/pull/3

### DIFF
--- a/tasks/run.js
+++ b/tasks/run.js
@@ -129,14 +129,14 @@ function runRouter(opts) {
 		});
 
 	return configureAndSpawn(envVars, function (env) {
-		var bin = opts.https ? 'next-router-https' : 'next-router';
+		var bin = opts.https ? `${opts.router}-https` : `${opts.router}`;
 		return [bin, { env: env }];
 	});
 }
 
-function ensureRouterInstall() {
-	return exec('which next-router')
-		.catch(function () { throw new Error('You need to install the next router first!  See docs here: https://github.com/Financial-Times/next-router'); });
+function ensureRouterInstall(router) {
+	return exec(`which ${router}`)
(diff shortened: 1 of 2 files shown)


tasks/run.js
```diff
@@ -196,6 +196,7 @@ module.exports = function (program, utils) {
 		.option('--key <file>', 'Specify a certificate key to use with HTTPS. Use with --https.')
 		.option('--local-apps <apps>', 'Specify extra apps that are running locally, as comma-seperated `[name]=[port]`, e.g. `service-worker=3001,front-page=3002`')
 		.option('-p --port <port>', 'Port to run the router through', 5050)
+		.option('--router <router>', 'Router to run using', 'next-router')
```
wheresrhys: or should this be just `next`?

Pro - more intuitive to type: `--router next` or `--router kat`
Con - less flexible - can't have a router that doesn't end in `-router`... but then again, why would you?
Con - a bit more magic

tasks/run.js
```diff
@@ -129,14 +129,14 @@ function runRouter(opts) {
 		});
 
 	return configureAndSpawn(envVars, function (env) {
-		var bin = opts.https ? 'next-router-https' : 'next-router';
+		var bin = opts.https ? `${opts.router}-https` : `${opts.router}`;
```
dapenguin: Why is `${opts.router}` inside back-ticks for the last bit of the statement? Could just refer to `opts.router`.

tasks/run.js
dapenguin: I'm easy on this one, but think the first con listed is a good enough reason to just allow us to specify the full name of the router.

leggsimon: I'm easy either way

PR #476, opened 2017-09-04
normalise app name in ship command

leggsimon:
If an app name (in it's package.json) wasn't prefixed with `ft-next` it meant it wasn't getting stripped out, this meant that the `kat-api` wasn't being scaled up properly 

 🐿 v2.5.16


adgad:
Seems legit 👍

PR #479, opened 2017-09-13
Update foreman to 2.0.0

sjparkinson:
`foreman@1.4.1` depends on `shell-quote@1.4.3` which contains a known vulnerability.

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,7 +17,7 @@
     "directly": "^1.0.1",
     "dotenv": "^1.2.0",
     "fetchres": "^1.0.4",
-    "foreman": "^1.4.1",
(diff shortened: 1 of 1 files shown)


sjparkinson:
@wheresrhys any thoughts on this update?

I'll dig into the changes.

wheresrhys:
foreman probably shouldn't be a dependency. What's it used for?

sjparkinson:
https://github.com/Financial-Times/n-heroku-tools/blob/47e3b1c9649b0618ab58eec93a1118e16e8a02eb/tasks/run.js#L105-L109

So it's all about the file watching and restarting express on changes.

PR #484, opened 2017-10-25
Provision to the ft-customer-products org 

adambraimbridge:
Provision to the ft-customer-products org rather than the (old) financial-times org.

 🐿 v2.5.16


alexnaish:
Fixed by #486

PR #486, opened 2017-12-12
Update and fix drydock / provisioning

alexnaish:
🐿 v2.5.16


circle.yml
```diff
@@ -1,6 +1,6 @@
 machine:
   node:
-    version: 4.1.1
+    version: 8.9.3
```
sjparkinson: <3

PR #490, opened 2018-01-16
Fix for issue with hyphenated environment variables

samuele-artuso:
Make `n-heroku-tools` set `UPPERCASE_SNAKECASE` style names for those env vars that hold the port number of apps that need to be run locally.

This is a fix for https://github.com/Financial-Times/n-heroku-tools/issues/489, where the rationale is also explained.

The old `lowercase-spinalcase` style env vars are still being set for backward compatibility.

Related PR: https://github.com/Financial-Times/next-router/pull/128.

### DIFF
--- a/tasks/run.js
+++ b/tasks/run.js
@@ -128,7 +128,7 @@ function runRouter (opts) {
 	(opts.localApps || [])
 		.concat({ name: normalizeName(packageJson.name, { version: false }), port: opts.localPort })
 		.forEach(function (localApp) {
-			const localAppEnvVar = localApp.name.toUpperCase().replace(/\-/, '_');
+			const localAppEnvVar = localApp.name.toUpperCase().replace(/\-/g, '_');
 			envVars[localAppEnvVar] = localApp.port;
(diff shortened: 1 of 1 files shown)


tasks/run.js
```diff
@@ -128,12 +128,14 @@ function runRouter (opts) {
 	(opts.localApps || [])
 		.concat({ name: normalizeName(packageJson.name, { version: false }), port: opts.localPort })
 		.forEach(function (localApp) {
+			const localAppEnvVar = localApp.name.toUpperCase().replace(/\-/, '_');
```
bjfletcher: `'email-article-api'.replace(/-/, '_')` yields `email_article-api`... you'll want the `/g` flag to replace all occurrences of the hyphen

bjfletcher:
a comment to explain why in the code will help (a brief summary from your issue), so coders new to this code will understand why we're doing this

tasks/run.js
samuele-artuso: I had missed that! Thanks!

PR #493, opened 2018-01-23
Use smoke tests from n-test

adgad:
For backwards compatibility, switch out the smoke tests in n-heroku-tools for n-test.

In an ideal world there wouldn't be a dependency here. I think local smoke tests can start using n-test directly. The deploy checks will be harder to unpick, so will create an issue to mull over.


tasks/deploy.js
```diff
@@ -87,7 +91,7 @@ module.exports = function (program, utils) {
 		.command('deploy [app]')
 		.description('runs haikro deployment scripts with sensible defaults for Next projects')
 		.option('-s, --skip-gtg', 'skip the good-to-go HTTP check')
-		.option('-a, --authenticated-smoke-tests', 'authenticate smoke tests with a backend authorization key')
+		.option('-a, --authenticated-moke-tests', 'authenticate smoke tests with a backend authorization key')
```
i-like-robots: moke

sjparkinson:
Any chance your changes mean we can look to remove `nightwatch`?

It's the cause of the whitesource security things, it depends on the broken ejs package 😢 .

lib/smoke-test.js
```diff
@@ -1,277 +0,0 @@
-'use strict';
```
wheresrhys: OMG, I'm getting all emotional seeing this file die :cry:

adgad: Death is just another path, one that we all must take. The grey rain-curtain of this world rolls back, and all turns to silver glass, and then you see it.

adgad: (sorry I was watching Lord of the Rings the other day)

wheresrhys: Would you like to borrow my scythe and cowl?

adgad:
@sjparkinson Not quite yet, because this doesn't do cross browser.

But I did create this - https://github.com/Financial-Times/n-test/issues/10 - would like to have another command that takes the same /test/smoke.js but run against cross browser tests. @rowanbeentje mentioned the apps team use http://webdriver.io/ so could look into that.

i-like-robots:
The name "Webdriver" gives me nightmares.

rowanbeentje:
😀 WebDriver has so many problems, but webdriver.io tries hard to paper them over - it has a nice promise-based API, and will handle a lot of the "stale element reference" type errors for you automatically.

PR #494, opened 2018-01-23
Pass build params

leggsimon:
Adds ability to pass build parameters to the triggered builds. The reason I'm wanting to add this is so that we can report on user-facing apps that fail on an n-ui tagged release.

With this change, an n-ui tagged release would now run:
```sh
nht rebuild --all --serves user-page --params NUI_RELEASE_BUILD=true
```

Then `next-webhooks` can filter and failed builds with the `NUI_RELEASE_BUILD` build_parameter and notifiy the person that released n-ui so that they can rebuild.

I'm wary this may be contentious so, thoughts?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,5 @@ devs.json
 npm-debug.log
 .eslintrc.js
 .editorconfig
+*.env*
+.vscode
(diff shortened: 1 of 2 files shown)


leggsimon:
I would actually still like this feature

ker-an:
Haha no problem, I will open it back up. I was being ruthless in my cleanup of the repos SEiT owns.

PR #496, opened 2018-02-07
Move URL resolving for smoke tests into here

adgad:
🐿 v2.7.0

In an attempt to make the smoke testing more generic/open-sourceable, move the bit where we resolve the heroku URL into here (where it makes more sense, being n-heroku-tools 😬 )


i-like-robots:
👍

PR #497, opened 2018-02-19
FT-specific n-test stuff

adgad:
Attempt to make n-test less FT.com specific, by moving the FT.com specific stuff here.

* Move the cache headers test into here
* Replace --auth with a global FT_NEXT_BACKEND_KEY header

### DIFF
--- a/lib/verify-cache-headers.js
+++ b/lib/verify-cache-headers.js
@@ -1,0 +1,60 @@
+const verifyCacheHeaders = (headers, path) => {
+
+	const cacheErrors = [];
+
+	if (!headers['surrogate-control'] && !headers['cache-control']) {
+		cacheErrors.push(`Each ${path} should specify a Cache-Control and/or a Surrogate-Control header`);
+	}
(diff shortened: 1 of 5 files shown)


lib/verify-cache-headers.js
```diff
@@ -0,0 +1,60 @@
...
+	} catch(errors) {
+		okay = false;
+		problems = errors;
+	}
+	return {
+		expected: 'Cache-Control headers should be sensible',
+		actual: okay || problems,
+		result: okay
+	};
+};
```
i-like-robots: I wonder if this should be a future clean up to use [assert](https://nodejs.org/api/assert.html) or some other expectations to make this more readable?

i-like-robots:
Nice one 👍

lib/verify-cache-headers.js
adgad: Potentially. Main reason I didn't end up going with assert is that it only logs errors, so would have had to write something to show useful success logs anyway.

PR #502, opened 2018-04-09
Support fetching session tokens when configuring the environment vars

tbergmen:
🐿 v2.7.0

### DIFF
--- a/tasks/configure.js
+++ b/tasks/configure.js
@@ -44,8 +44,35 @@ function fetchFromVault (source, target, serviceData) {
 			}, {});
 
 			return Object.assign({}, shared, appVars.data);
+		})
+		.then((keys) => {
+			const { TEST_USER_TYPES, TEST_SESSIONS_URL, TEST_SESSIONS_API_KEY } = keys;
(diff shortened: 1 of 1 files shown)


tasks/configure.js
```diff
@@ -44,9 +44,36 @@ function fetchFromVault (source, target, serviceData) {
...
+			return Promise.all(userTypes.map((type) => appendSessionToken(keys, type.trim().toLowerCase(), TEST_SESSIONS_URL, TEST_SESSIONS_API_KEY)))
+				.then(() => keys)
+				.catch((e) => {
+					console.log('Couldn\'t fetch the session tokens. Please check TEST_USER_TYPES, TEST_SESSIONS_URL and TEST_SESSIONS_API_KEY environment variables.'); // eslint-disable-line no-console
+					console.error(e); // eslint-disable-line no-console
+				});
 		});
 }
 
+const appendSessionToken = (keys, userType, url, apiKey) => {
```
adgad: might be nicer if this function didn't mutate keys, but returned something that gets added to keys while it's in scope above?

PR #503, opened 2018-04-11
Update Heroku error to make it clearer when build fails

ker-an:
🐿 v2.8.0

image

When a test fails, the build proceeds to `make tidy` and tries to clean up a Heroku app that doesn't exist. The resulting error is quite confusing (see image above) so this hopefully makes it a bit more clear!

### DIFF
--- a/tasks/destroy.js
+++ b/tasks/destroy.js
@@ -5,18 +5,22 @@ let spawn = require('shellpromise');
 function task (options) {
 	let app = options.app;
 	let verbose = options.verbose;
-	let promise = Promise.resolve();
-	if (verbose) {
-		promise = promise.then(function () {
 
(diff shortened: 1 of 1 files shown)


tasks/destroy.js
```diff
@@ -5,18 +5,22 @@ let spawn = require('shellpromise');
 function task (options) {
 	let app = options.app;
 	let verbose = options.verbose;
-	let promise = Promise.resolve();
-	if (verbose) {
-		promise = promise.then(function () {
 
+	return spawn('heroku info ' + app, {verbose: true }).then(function () {
```
adgad: Do we still need `verbose` here now, or can we get rid of it?

ker-an: No, we don't! I'll get rid of it.

PR #508, opened 2018-05-24
Tidy up deprecated stuff

adgad:
* Remove bottle command, which is no longer used
* Remove old smoke docs
* Remove empty a11y file
* Remove deprecated --no-splunk option for configure
* Remove `next-build-tools` as a binary (had to keep nbt as it's still being used in places)

This should be releasable as a minor version, as nothing is using the deleted the things 🤞


tasks/bottle.js
```diff
@@ -1,299 +0,0 @@
-'use strict';
-
-const shell = require('shellpromise');
```
wheresrhys: :cry:

scripts/generate-docs.sh
```diff
@@ -1,2 +1,2 @@
...
+printf "# n-heroku-tools\nThis library is a command line tool that orchestrates [Heroku](https://www.heroku.com/) and [Amazon S3](https://aws.amazon.com/s3/) deployments for [Next](https://github.com/Financial-Times/next/wiki), based on configuration in the [Next service registry](https://next-registry.ft.com/v2/) and [Vault](https://www.vaultproject.io/).\n### Installation\nIn order to use this tool, run\n\`\`\`\nnpm install @financial-times/n-heroku-tools --save-dev\n\`\`\`\n\n### Development\n ### Usage\nIn order to use \`n-heroku-tools\` the following commands are available in your command line:\n`./bin/n-heroku-tools.js`\n\n*Note*: The README.md is automatically generated.  Run \`make docs\` to update it.\n"
```
gvonkoss: If there are no development instructions, maybe get rid of the heading?

PR #509, opened 2018-06-13
Update vulnerable versions of dependency modules.

roland-vachter:
🐿 v2.9.0

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,7 +19,7 @@
     "denodeify": "^1.2.0",
     "dotenv": "^1.2.0",
     "fetchres": "^1.0.4",
-    "foreman": "^1.4.1",
+    "foreman": "^3.0.0",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -39,7 +39,7 @@
     "express": "^4.13.3",
     "fetch-mock": "^3.0.0",
     "lintspaces-cli": "^0.1.1",
-    "mocha": "^2.2.4",
+    "mocha": "^5.0.0",
```
adgad: Do you now need to add `--exit` after the `mocha` command?

roland-vachter: --exit                                  force shutdown of the event loop after test run: mocha will call process.exit

roland-vachter: Normally you shouldn't use --exit as far as I understand.

PR #511, opened 2018-07-25
serviceData has `system.code`s, not `system.systemCode`s

adambraimbridge:
See: https://circleci.com/gh/Financial-Times/rich/107

image


simonplend:
I was pondering this last night. What about if we removed the check against system code entirely? It's been broken for a long time and I think it would simplify things if we only did lookup based on name.

PR #516, opened 2018-08-15
Support https; Nicer errors; Moar emojis; Reject on any error other 'FetchError: network timeout'

adambraimbridge:
Looking good to go.

### DIFF
--- a/lib/wait-for-ok.js
+++ b/lib/wait-for-ok.js
@@ -18,12 +18,12 @@ module.exports = function (url) {
 					}
 				})
 				.catch(error => {
-					//if (error.includes('FetchError: network timeout')) {
-					//	console.log(`👋 Hey, ${url} doesn't seem to be responding yet, so there's that. You're amazing, by the way. I don't say that often enough. But you really are.`); // eslint-disable-line no-console
-					//}
-					//else {
+					if (error.indexOf('FetchError: network timeout') > -1) {
+						console.log(`👋 Hey, ${url} doesn't seem to be responding yet, so there's that. You're amazing, by the way. I don't say that often enough. But you really are.`); // eslint-disable-line no-console
+					}
+					else {
 						console.log(`😿 ${url} Error: ${error}`); // eslint-disable-line no-console
-					//}
+					}
 				});
 		}
 		checker = setInterval(checkGtg, 3000);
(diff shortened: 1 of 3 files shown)


adambraimbridge:
Seems to be okay ... 
image

adambraimbridge:
If there are no errors: It floats!

image

adambraimbridge:
Update:
 - [ ] Moar emojis! 😻
 - [ ] If there's a "real" error, don't wait two minutes to time out.
 - [ ] De-duplication of final error message.

tasks/float.js
```diff
@@ -60,7 +60,6 @@ function task (opts) {
 		log.art.canoe();
 		log.success('IT FLOATS!');
 	}).catch(function (err) {
-		log.error('Man overboard', err);
```
simonplend: :+1:

lib/wait-for-ok.js
```diff
@@ -5,23 +5,31 @@ module.exports = function (url) {
 		let timeout;
 		let checker;
 		function checkGtg () {
-			console.log(`polling: ${url}`); // eslint-disable-line no-console
+			console.log(`⏳ polling: ${url}`); // eslint-disable-line no-console
 			fetch(url, { timeout: 2000, follow: 0 })
 				.then(function (response) {
 					if (response.ok) {
-						console.log(`poll ${url} ok`); // eslint-disable-line no-console
+						console.log(`⌛️ ${url} ok!`); // eslint-disable-line no-console
```
simonplend: How about a ✔️ emoji when it's ok...

lib/wait-for-ok.js
```diff
@@ -5,23 +5,31 @@ module.exports = function (url) {
...
 						clearTimeout(timeout);
 						clearInterval(checker);
 						resolve();
 					} else {
-						console.log(`poll ${url} not ok`); // eslint-disable-line no-console
+						console.log(`⌛️ ${url} not ok`); // eslint-disable-line no-console
```
simonplend: ...and a ❌ for when they're not.

I figure these are easy to scan for in output to determine success or failure and similar to the output that test runners like mocha and n-test provide.

:thinking:

lib/wait-for-ok.js
```diff
@@ -5,23 +5,31 @@ module.exports = function (url) {
...
+					}
+				})
+				.catch(error => {
+					if (error.type && error.type === 'request-timeout') {
+						console.log(`👋 Hey, ${url} doesn't seem to be responding yet, so there's that. You're amazing, by the way. I don't say that often enough. But you really are.`); // eslint-disable-line no-console
+					}
+					else {
+						reject(`😿 ${url} Error: ${error}`);
+						clearInterval(checker);
```
simonplend: Good :cat:ch!

PR #524, opened 2018-10-08
Upgrade next-metrics to v3.x.x

simonplend:
https://github.com/Financial-Times/next/issues/170


simonplend:
Given the amount of yak shaving required to make this work and the fact that nobody (to our knowledge) uses these metrics, we're going to remove the sending of metrics from `nht`. It is likely we'll find a better way to do this as part of our work around improving how we use Heroku / through the Core UI project.

simonplend:
Superseded by #525.

PR #525, opened 2018-10-08
Remove asset size metrics from deploy-hashed-assets / deploy-static

simonplend:
This is tricky to do now that FT Graphite v2 requires every client that is sending metrics to have an app UUID. Given that `nht` is a CLI tool that has been sending metrics from CI environments, this starts to get a bit messy.

We've decided to drop the sending of metrics from `nht` as there is ongoing work to improve how we deploy and how we build the UI for applications. Hopefully those efforts will be able to reintroduce asset size metrics in some other form.

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,7 +25,6 @@
     "isomorphic-fetch": "^2.0.0",
     "md5-file": "^3.1.0",
     "mime": "^1.3.4",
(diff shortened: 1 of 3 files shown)


simonplend:
Related to https://github.com/Financial-Times/next/issues/170

PR #526, opened 2018-10-17
Add `nht review-app` task

taktran:
Outputs the name of the review app if successful.

Created for https://github.com/Financial-Times/next/issues/212

## TODO

* [x] Test it works on https://github.com/Financial-Times/next-search-page/tree/post-build - https://circleci.com/gh/Financial-Times/next-search-page/1634
* [x] Delete review app, if it already exists
* [x] ~~Remove heroku app when provision is complete~~ - do this outside of the task

### DIFF
--- a/lib/pipelines.js
+++ b/lib/pipelines.js
@@ -2,50 +2,41 @@
 const herokuAuthToken = require('./heroku-auth-token');
 const co = require('co');
 const shellpromise = require('shellpromise');
+const api = require('./heroku-api');
 const spawn = require('child_process').spawn;
 
-function api (url, token){
-	return fetch(
-		'https://api.heroku.com' + url,
-		{
-			headers: {
-				'Accept': 'Accept: application/vnd.heroku+json; version=3',
-				'Content-Type': 'application/json',
-				'Authorization': 'Bearer ' + token
-			}
-		}
-	).then(response => {
-		if(!response.ok){
-			let err = new Error(`BadResponse: ${response.status} ${response.statusText}`);
-			err.name = 'BAD_RESPONSE';
-			err.status = response.status;
-			throw err;
-		}
-
-		return response.json();
-	});
-}
-
 async function info (pipelineName){
-	const token = await herokuAuthToken();
-	const pipeline = await api(`/pipelines/${pipelineName}`, token);
+	const authToken = await herokuAuthToken();
+	const pipeline = await api({
+		endpoint: `/pipelines/${pipelineName}`,
+		authToken
+	});
 
 	return pipeline;
 }
 
 function getApps (pipelineName){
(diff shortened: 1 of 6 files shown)


tasks/review-app.js
```diff
@@ -0,0 +1,173 @@
...
+			return {
+				...defaultHeaders,
+				Authorization: `Bearer ${key}`
+			};
+		});
+}
+
+const throwIfNon2xx = async res => {
+	const { status, url } = res;
+	if (!status || !status.toString().match(/^2.*/)) {
```
adgad: could check `res.ok` here rather than checking the status

tasks/review-app.js
```diff
@@ -0,0 +1,173 @@
...
+}
+
+/**
+* Assume
+* 	* app is VAULT_SOURCE, and is package.json name (could assume it's the package.json name, like `nht configure`)
+*/
+module.exports = function (program) {
+	program
+		.command('review-app [app]')
+		.description('Create a heroku review app and print out the app name created')
```
adgad: Perhaps this should say something like "Create or retrieve a heroku review app for a given branch, and print out the review app name" - to make it explicit that it won't recreate.

taktran: I think the redeploy was only for the normal heroku review app, for the review apps beta, you need to explicitly delete before you can build again. I just tried it with different commits (even if it is a subsequent commit in the branch)

I'm gonna add a bit to delete it first if it already exists.

adgad: Hmm. Is that true even if there is a PR open? I can _swear_ that it seemed to rebuild when we tried it 🤔

taktran: Yep, there's a PR here: https://github.com/Financial-Times/next-search-page/pull/158

But it might be something up with my test environment. To test properly, I'm going to need to run it on CI. TBC.

lib/pipelines.js
```diff
@@ -2,43 +2,41 @@
...
+
+	return pipeline;
 }
 
 function getApps (pipelineName){
 	return co(function* (){
-		let token = yield herokuAuthToken();
-		let pipelines = yield api('/pipelines', token);
+		let authToken = yield herokuAuthToken();
+		let pipelines = yield api({
```
adgad: Doesn't need to be part of this PR - but we should consider refactoring this to use `async/await`.

tasks/review-app.js
taktran: Fixed it so that if there is an existing review app, it deletes it first and prints out `Review app already created for post-build branch. Deleting existing review app first.`, then creates a new review app.

You can see it running here: https://circleci.com/gh/Financial-Times/next-search-page/1634

PR #527, opened 2018-10-18
Stop scaling the production apps when shipping

adgad:
🐿 v2.7.0

## What
Stop trying to scale production apps when deploying.

## Why

Heroku have changed something in their stack, so that when preboot is enabled (as it is and needs to be on all of our apps), trying to scale just after deploying will throw an error. This only happens if the current production formation is different to the specified formation in the service registry.

This leads to confusing behaviour in CI, where the deploy step will go red - even though the app has been deployed.

This resolves https://github.com/Financial-Times/next/issues/150

We have also discussed in https://github.com/Financial-Times/next/issues/182 whether keeping scaling information in code is even necessary. (tl;dr - no)

## Next Steps
- [ ] Remove all the formation information from next-service-registry

### DIFF
--- a/tasks/ship.js
+++ b/tasks/ship.js
@@ -86,33 +86,6 @@ function task (opts) {
 		log.info('Promote slug to production');
 		yield pipelines.promote(apps.staging);
 		log.success('Slug promoted');
-		if(opts.scale){
-			log.log('scale enabled');
-			let source = appName;
-			let scaleTasks = [
-				scale({
-					source:source,
-					target:apps.staging,
-					registry: REGISTRY_URI
-				}),
-				scale({
-					source:source,
-					target:apps.production.eu,
-					registry: REGISTRY_URI
-				})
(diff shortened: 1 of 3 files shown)


kitkat119:
@adgad I don't get what that was doing in the first place? Why was it trying to scale staging and prod after they had already been deployed? Was the service reg specifying different number of dynos for some apps?

adgad:
@taktran Yeah I wasn't sure about that. I'll get rid of it.

@kitkat119 Excellent question. So there's two bits of scaling going on:

1) scale up the staging app before testing, then scale down after finishing. This is I guess to save money since we don't need staging all the time, so we should keep this.

2) scaling the production apps to whatever service registry says. Most of the time the apps should already be the same, so scaling won't do anything. I guess the main purpose of this was so whenever we manually change the formation of an app, it will scale back to the "defined" level when we deploy.

PR #529, opened 2018-10-31
Use delete app api rather than delete review app api  🐿 v2.10.3

taktran:
Delete app api seems to be more thorough than delete review app.

Sometimes a review app can be stuck in a `deleting` state, and even sending further requests to delete the review app don't do anything. Deleting with the app api seems to work all the time though.

Some example data of the `deleting` limbo state: https://financialtimes.slack.com/archives/CD68S13LP/p1540981485044000

### DIFF
--- a/tasks/review-app.js
+++ b/tasks/review-app.js
@@ -122,7 +122,7 @@ const getAppName = async (appId) => {
 		});
 };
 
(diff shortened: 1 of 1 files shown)


taktran:
This is an updated way to deal with the deleting issue: https://github.com/Financial-Times/next/issues/230

PR #530, opened 2018-11-02
Enable review apps to be configured on the pipeline rather than the app

kitkat119:
🐿 v2.10.3

We wanted to be able to set our config vars on the pipeline rather than having to wait for the name of the spun up review app. Fortunately Heroku let us know about a prototype API that allowed us to do just that. Unfortunately this api is subject to change but they should give us a week's notice if this is going to happen!

Initially I tried using this prototype api to configure all stages of the pipeline but it turned that it currently only supports the 'test' stage, so I have had to keep the standard heroku api, which sets config vars directly on the staging and prod apps.

Because of this, I was going to write a separate nht task that would specifically configure review apps but I figured there wasn't a lot that would be different and so I have had a go at including it in the one configure task. If it's a bit hacky and you reckon it would be clearer in it's own nht task, I will change it.

To configure review apps on the pipeline, set the target in the command to 'review-app' e.g. `nht configure ft-next-search-page review-app`. Otherwise the command remains the same for configuring apps directly, e.g. `nht configure ft-next-search-page ft-next-search-page-eu`.

All feedback very welcome :)

### DIFF
--- a/lib/heroku-auth-token.js
+++ b/lib/heroku-auth-token.js
@@ -1,21 +1,35 @@
 'use strict';
 
 let shellpromise = require('shellpromise');
+let authToken;
 
-module.exports = function () {
+const getAuthFromCli = () => {
+	return shellpromise('heroku auth:whoami 2>/dev/null')
+		.then(function () {
+			return shellpromise('heroku auth:token 2>/dev/null');
+		})
+		.then(function (token) {
+			return token.trim();
+		})
+		.catch(function (err) {
+			console.error(err); // eslint-disable-line no-console
+			throw new Error('Please make sure the Heroku CLI is authenticated by running `heroku auth:token`');
+		});
+};
+
+module.exports = async function () {
 	if (process.env.HEROKU_AUTH_TOKEN) {
+
 		return Promise.resolve(process.env.HEROKU_AUTH_TOKEN);
+
 	} else {
-		return shellpromise('heroku auth:whoami 2>/dev/null')
-			.then(function () {
-				return shellpromise('heroku auth:token 2>/dev/null');
-			})
-			.then(function (token) {
-				return token.trim();
-			})
-			.catch(function (err) {
-				console.error(err); // eslint-disable-line no-console
-				throw new Error('Please make sure the Heroku CLI is authenticated by running `heroku auth:token`');
-			});
+
+		if (authToken) {
+			return authToken;
+		} else {
+			authToken = await getAuthFromCli();
+			return authToken;
+		}
+
 	}
 };
--- a/lib/heroku-config-vars.js
+++ b/lib/heroku-config-vars.js
@@ -9,7 +9,7 @@ class HerokuConfigVars {
 
 		this.headers = {};
 		this.endpoint = null;
-	
+
 		if (target === 'review-app') {
 			this.headers.Accept = 'application/vnd.heroku+json; version=3.pipelines';
 			this.endpoint = `/pipelines/${pipelineId}/stage/review/config-vars`;
@@ -33,7 +33,7 @@ class HerokuConfigVars {
 	async makeApiCall (options = {}) {
 
 		let herokuApiResponse;
-	
+
 		try {
 			herokuApiResponse = await herokuApi({
 				endpoint: this.endpoint,
@@ -50,7 +50,7 @@ class HerokuConfigVars {
 				throw err;
 			}
 		}
-	
+
 		return herokuApiResponse;
 	}
 
--- a/lib/heroku-config-vars.js
+++ b/lib/heroku-config-vars.js
@@ -1,0 +1,59 @@
+const herokuApi = require('./heroku-api');
+
+class HerokuConfigVars {
+
+	constructor (settings) {
+		const { target, pipelineId, authToken } = settings;
+
+		this.authToken = authToken;
+
+		this.headers = {};
+		this.endpoint = null;
+	
+		if (target === 'review-app') {
+			this.headers.Accept = 'application/vnd.heroku+json; version=3.pipelines';
+			this.endpoint = `/pipelines/${pipelineId}/stage/review/config-vars`;
+		} else {
+			this.headers.Accept = 'application/vnd.heroku+json; version=3';
+			this.endpoint = `/apps/${target}/config-vars`;
+		}
+	}
+
+	get () {
+		return this.makeApiCall();
+	}
+
+	set (patch) {
+		return this.makeApiCall({
+			method: 'patch',
+			body: JSON.stringify(patch)
+		});
+	}
+
+	async makeApiCall (options = {}) {
+
+		let herokuApiResponse;
+	
+		try {
+			herokuApiResponse = await herokuApi({
+				endpoint: this.endpoint,
+				authToken: this.authToken,
+				options: {
+					headers: this.headers,
+					...options
+				}
+			});
+		} catch(err) {
+			if (err.name === 'BAD_RESPONSE' && err.status === 404) {
+				throw new Error('The specified app does not seem to exist in Heroku');
+			} else {
+				throw err;
+			}
+		}
+	
+		return herokuApiResponse;
+	}
+
+}
+
+module.exports = HerokuConfigVars;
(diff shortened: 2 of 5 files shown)


adgad:
https://devcenter.heroku.com/articles/review-apps-beta#pipeline-config-vars

Link to the docs for the prototype API

tasks/configure.js
```diff
@@ -97,18 +105,23 @@ function task (opts) {
 		});
 	}
 
-	let authorizedPostHeaders = {
-		'Accept': 'application/vnd.heroku+json; version=3',
+	const acceptHeader = target === 'review-app' ? 'application/vnd.heroku+json; version=3.pipelines' : 'application/vnd.heroku+json; version=3';
```
adgad: Given we have both the URL and the header to change, would it perhaps be better to combine both in the same conditional:

```
let fetchConfig;

if(target === 'review-apps') {
   fetchConfig = fetch( reviewAppUrl, { headers});
} else {
   fetchConfig = fetch( regularUrl, { headers});
}

...

Promise.all([
   blah,
   fetchConfig,
   blah
]);

```

adgad: (not necessarily an if statement like that - but basically only having one conditional the decides both at the same time.)

lib/heroku-config-vars.js
```diff
@@ -0,0 +1,59 @@
...
+		this.authToken = authToken;
+
+		this.headers = {};
+		this.endpoint = null;
+
+		if (target === 'review-app') {
+			this.headers.Accept = 'application/vnd.heroku+json; version=3.pipelines';
+			this.endpoint = `/pipelines/${pipelineId}/stage/review/config-vars`;
+		} else {
+			this.headers.Accept = 'application/vnd.heroku+json; version=3';
```
simonplend: This line isn't needed as it's the default `Accept` header in `herokuApi`.

tasks/configure.js
```diff
@@ -1,33 +1,37 @@
...
+const normalizeName = require('../lib/normalize-name');
+const vault = require('../lib/vault');
+const pipelines = require('../lib/pipelines');
+const HerokuConfigVars = require('../lib/heroku-config-vars');
 
 const FORBIDDEN_ATTACHMENT_VARIABLES = [
 	'DATABASE_URL'
 ];
 
 const DEFAULT_REGISTRY_URI = 'https://next-registry.ft.com/v2/';
```
adgad: (separate to this PR)

Should we remove the ability to change the registry URL? It's legacy from the last time we migrated registries. AFAIK everything can/should just use the v2 one now - not sure whether it's ever going to be useful again later.

Thoughts @sjparkinson ?

tasks/configure.js
```diff
@@ -1,33 +1,37 @@
...
-		const serviceData = findService(json, normalizeName(source));
-		if (!serviceData) {
-			throw new Error('Could not find a service in the registry, with `name` or `systemCode`, matching ' + source + '. Please check the service registry.');
-			return false;
-		}
-		else {
-			return serviceData;
-		}
-	});
+const getServiceData = (source, registry) => {
```
adgad: this could be `(source, registry) => fetch(registry)...` as it's just returning straight away.

tasks/configure.js
```diff
@@ -1,33 +1,37 @@
...
+			}
+			else {
+				return serviceData;
+			}
+		});
+};
 
-function fetchFromVault (source, target, serviceData) {
-	const path = serviceData.config.replace('https://vault.in.ft.com/v1/','');
+function fetchFromVault (serviceData) {
```
adgad: Should we convert this to an arrow function as well, for consistency with the rest of the file?

Maybe even async?

lib/heroku-config-vars.js
```diff
@@ -0,0 +1,59 @@
+const herokuApi = require('./heroku-api');
+
+class HerokuConfigVars {
+
+	constructor (settings) {
```
simonplend: This would really benefit from a comment before the method that explains why we're handling the `review-app` target differently to others, including a mention that we're using a preview API and a link to any relevant Heroku docs.

JSDoc style comments are the general convention for JavaScript i.e. http://usejsdoc.org/howto-es2015-classes.html

tasks/configure.js
```diff
@@ -97,76 +106,62 @@ function task (opts) {
...
-				})
-				.then(response => {
-					if (response.status !== 200) return response.json().then(({id, message}) => Promise.reject(new Error(`Heroku Error - id: ${id}, message: ${message}`)));
-					console.log(target + ' config vars are set'); // eslint-disable-line no-console
-				}));
-			});
+	console.log('Retrieving pipeline details from Heroku...'); // eslint-disable-line no-console
+
+	// TODO: Investigate what herokuAuthToken is doing
+	const [ authToken, pipelineId ] = await Promise.all([
```
simonplend: Reminder to check if we can use the `herokuAuthToken` function more efficiently. Is it currently making redundant API calls using the heroku CLI?

tasks/configure.js
```diff
@@ -97,76 +106,62 @@ function task (opts) {
...
+
+	Object.keys(patch).forEach(function (key) {
+		if (patch[key] === null) {
+			console.log(`Deleting config var: ${key}`); // eslint-disable-line no-console
+		} else if (patch[key] !== current[key]) {
+			console.log(`Setting config var: ${key}`); // eslint-disable-line no-console
+		}
+	});
+
+	console.log('Setting config vars', Object.keys(patch)); // eslint-disable-line no-console
```
adgad: Is this duplicating the logging above? 🤔

tasks/configure.js
kitkat119: Yep, I will change it! We didn't change it to async yet as felt it would be worth writing tests first...don't know which is scarier though

taktran:
I noticed on [ft-next-search-page](https://dashboard.heroku.com/pipelines/3a0edcd8-cf12-4122-93da-c4b28ebd25cb) that there are the following errors when deploying:

> Could not create review app. Invalid app.json. Config var "AWS_ACCESS_HASHED_ASSETS" is required. Config var "AWS_SECRET_HASHED_ASSETS" is required.
 
screen shot 2018-11-07 at 1 06 46 pm

It's because `AWS_ACCESS_HASHED_ASSETS` and `AWS_SECRET_HASHED_ASSETS` are lower case in vault. To fix this, I updated `app.json` to use the lowercase versions.

Just a comment in case anyone encounters this in future when configuring apps with secrets.

tasks/configure.js
kitkat119: I thought it's ok as we log what is happening with:
```
Retrieving pipeline details from Heroku...
Retrieving current and desired config vars...
```
and then this signals that they are being set and it prints out every var that it is setting.

tasks/configure.js
sjparkinson: I believe this is so that Enterprise Tools can access their respective registries?

tasks/configure.js
kitkat119: herokuAuthToken is getting the logged in user and then their token using the CLI. It is currently being called twice in configure.js as lib/pipelines calls it aswell, so will look into changing this.

tasks/configure.js
```diff
@@ -97,76 +103,62 @@ function task (opts) {
...
+
+	Object.keys(overrides).forEach(function (key) {
+		patch[key] = overrides[key];
+	});
+
+	Object.keys(patch).forEach(function (key) {
+		if (patch[key] === null) {
+			console.log(`Deleting config var: ${key}`); // eslint-disable-line no-console
+		} else if (patch[key] !== current[key]) {
+			console.log(`Adding new config var: ${key}`); // eslint-disable-line no-console
```
adgad: This would be updated values as well as new.

lib/heroku-auth-token.js
```diff
@@ -1,21 +1,35 @@
...
-			.then(function (token) {
-				return token.trim();
-			})
-			.catch(function (err) {
-				console.error(err); // eslint-disable-line no-console
-				throw new Error('Please make sure the Heroku CLI is authenticated by running `heroku auth:token`');
-			});
+
+		if (authToken) {
+			return authToken;
```
taktran: If you are going to store it, you could just use `process.env.HEROKU_AUTH_TOKEN`, and save on having another nested `if` statement ie,

    process.env.HEROKU_AUTH_TOKEN = await getAuthFromCli();

Setting `process.env.HEROKU_AUTH_TOKEN` inside this script wouldn't persist outside of this script being run too.

PR #531, opened 2018-11-07
Use existing review app if there is a conflict

taktran:
Rather than delete an existing review when a conflict is found, just use the same review app for builds.

Fixes https://github.com/Financial-Times/next/issues/230

Also replace mocha with jest and added unit tests for review app lib functions. Still want to do integration tests for the cli side, but can do that later.

CircleCI examples:

* [a new review app being created](https://circleci.com/gh/Financial-Times/next-search-page/1878)
* [reusing an existing review app](https://circleci.com/gh/Financial-Times/next-search-page/1887)


tasks/review-app.js
```diff
@@ -1,208 +1,59 @@
...
-
-	return createReviewApp()
+	return createReviewApp({ pipelineId, repoName, commit, branch, githubToken })
 		.then(res => {
 			const { status } = res;
 			if (status === 409) {
-				console.error(`Review app already created for ${branch} branch. Deleting existing review app first.`); // eslint-disable-line no-console
-				return deleteGitBranchApp({ pipelineId, branch, headers })
-					.then(createReviewApp);
+				console.error(`Review app already created for ${branch} branch. Using existing review app for build.`); // eslint-disable-line no-console
```
adgad: Maybe this should be a warn or info rather than error?

taktran: Nope, doesn't work with `warn` or `log` as it prints out to `stdin`, which I don't want since it needs to only print out the app name at the end.

eg, https://circleci.com/gh/Financial-Times/next-search-page/1922

The makefile thinks `TEST_APP=Review app already created for chores/nht-reuse-review-apps branch. Using existing review app for build. ft-next-sear-chores-nht-addywg` rather than `TEST_APP=ft-next-sear-chores-nht-addywg`

adgad: Ah I see. Maybe worth adding comment saying why it needs to be an error.

PR #553, opened 2019-02-19
Override default maximum size of HTTP headers

ker-an:
Node.js 8.14.0 introduced a maximum size for HTTP headers of 8KB. This ended up breaking a lot of our apps because our header sizes are much larger. In Node.js 8.15.0, you can specify a different maximum size using the `--max-http-header-size` option.

While adding a smoke test to next-router to test the maximum header size, we found that the test wasn't passing locally because of `nht run`. Turns out we need to add the `node --max-http-header-size=80000` option when starting node and that's what this PR does!

Resolves https://github.com/Financial-Times/n-heroku-tools/issues/552

### DIFF
--- a/tasks/run.js
+++ b/tasks/run.js
@@ -101,6 +101,16 @@ function runScript (opts) {
 		if (opts.subargs) {
 			args = args.concat(opts.subargs.replace(/^\[/, '').replace(/]$/, '').split(','));
 		}
+
+		/**
+		 * The default maximum size of HTTP headers is 8KB.
+		 * To override the default, we must pass in the --max-http-header-size option
+		 * and specify a maximum size.
+		 *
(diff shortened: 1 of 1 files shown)


ker-an:
This PR will break any apps that are pinned to Node.js v8.13.0. 

I am blocked from merging this until we resolve https://github.com/Financial-Times/next-transformations/issues/31.

tasks/run.js
```diff
@@ -101,6 +101,16 @@ function runScript (opts) {
 		if (opts.subargs) {
 			args = args.concat(opts.subargs.replace(/^\[/, '').replace(/]$/, '').split(','));
 		}
+
+		/**
+		 * The default maximum size of HTTP headers is 8KB.
```
adambraimbridge:
```suggestion
		 * Since node version 8.14, the default maximum size of HTTP headers is 8KB.
```

Issue #551, opened 2019-02-18
Make static assets S3 bucket name configurable


i-like-robots:
Currently the `deploy-hashed-assets` task will send files from the configured local directory (`./public/` by default) to a folder named `hashed-assets/app-name` (`app-name` being inferred from the application's `pkg.name`).

https://github.com/Financial-Times/n-heroku-tools/blob/master/tasks/deploy-hashed-assets.js#L66

With the Core UI project we intend to deploy all hashed assets to a shared folder. This is because we will be splitting our JS bundle into more granular parts with the intention that this will increase cache-ability, reduce the amount we bust the cache, and enable greater reuse of assets between our user facing applications.

i-like-robots:
Fixed by #556 and #560

PR #556, opened 2019-03-01
Make S3 paths configurable via CLI flag

oliverturner:
Currently, hashed assets are written to `/hashed-assets/${appName}/hashedName` where `appName` is read from `package.json`. This PR adds support for a command line flag `--app` to provide a custom value.

This will allow N-Gage to deploy the assets of Anvil-based apps to `/hashed-assets/anvil/hashedName` so that modules common across apps can be cached client-side.

This PR addresses issue #551. A companion PR to N-Gage is in flight.

 🐿 v2.12.0

### DIFF
--- a/lib/normalize-name.spec.js
+++ b/lib/normalize-name.spec.js
@@ -1,0 +1,26 @@
+var normalize = require("./normalize-name");
+
+describe("normalize", () => {
+  it("removes ft prefixes without options", () => {
+    var alpha = normalize("ft-alpha");
+    var beta = normalize("next-beta");
+    var gamma = normalize("@financial-times/gamma");
+    
+    expect(alpha).toBe("alpha");
+    expect(beta).toBe("beta");
+    expect(gamma).toBe("@financial-times/gamma");
+  });
+  it("removes ft prefixes and versions", () => {
+    var alpha = normalize("ft-alpha-v1", { version: false });
+    var beta = normalize("next-beta-v99", { version: false });
+    var gamma = normalize("ft-gamma-v123", { version: false });
+    var delta = normalize("next-delta-v123", { version: false });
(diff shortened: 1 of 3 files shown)


lib/normalize-name.spec.js
```diff
@@ -0,0 +1,26 @@
...
+    var delta = normalize("next-delta-v123", { version: false });
+    var epsilon = normalize("@financial-times/epsilon", { version: false });
+
+    expect(alpha).toBe("alpha-v1");
+    expect(beta).toBe("beta-v99");
+    expect(gamma).toBe("gamma");
+    expect(delta).toBe("delta");
+    expect(epsilon).toBe("epsilon");
+  });
+});
```
i-like-robots: Looks like test files are suffixed `.unit.test.js` in this repo

tasks/deploy-hashed-assets.js
```diff
@@ -7,6 +7,7 @@ const readFile = denodeify(require('fs').readFile);
 const waitForOk = require('../lib/wait-for-ok');
 const path = require('path');
 const aws = require('aws-sdk');
+const fs = require('fs');
```
i-like-robots: Unused dependency

i-like-robots:
Nice one! Makes sense to me.

lib/normalize-name.spec.js
```diff
@@ -0,0 +1,26 @@
+var normalize = require("./normalize-name");
+
+describe("normalize", () => {
```
taktran: I actually started a `add-tests` branch with a test for this with `describe.each`: https://github.com/Financial-Times/n-heroku-tools/blob/533857c744528889d25db2de8ccedde619767c5c/lib/normalize-name.test.js

PR #563, opened 2019-03-27
save response text/json on heroku errors

quarterto:
- try and parse it as json but don't use `response.json` up front in case it's not
- i don't know if this is safe ie i would hope heroku isn't echoing our sensitive data eg config vars in errors but who knows


quarterto:
yolo

PR #579, opened 2019-08-07
Add REDIS_URL to list of forbidden variables

jamesnicholls:
The REDIS_URL environment variables is set by the Heroku Redis add-on (https://devcenter.heroku.com/articles/heroku-redis#create-a-new-instance). Setting the variable in Vault will cause problems when the value changes.


andygout:
`next-myft-api` gets an env var from Vault called `REDIS_URL` but does not have its own Redis add-on (it shares an instance attached to another Heroku app).

Our solution is to simply rename `REDIS_URL` to something else (`REDIS_URL_REMOTE` or something better).

Let me know if there is a better solution.

PR #595, opened 2020-06-01
Update README to reference common usage of NHT

nickcolley:
As a newcomer it was confusing me if this repository was for `nht` which I see across the applications.

Default the repository to the common usecase while referencing the expanded name.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,10 +1,10 @@
-# n-heroku-tools
+# nht (n-heroku-tools)
 
 This library is a command line tool that orchestrates [Heroku](https://www.heroku.com/) and [Amazon S3](https://aws.amazon.com/s3/) deployments for [Next](https://github.com/Financial-Times/next/wiki), based on configuration in the [Next service registry](https://next-registry.ft.com/v2/) and [Vault](https://www.vaultproject.io/).
 
 ### Installation
 
-In order to use this tool, run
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,21 +1,21 @@
-# n-heroku-tools
+# nht (n-heroku-tools)
```
andygout, line 1: I would treat `n-heroku-tools` as the main name, as that is the repo name and its package name in the `npm` registry, and keep `(nht)` suffixed in parentheses more of a side note that it is also referred to as such.

I am generally against using initialisms/acronyms as the main identifier of things as they tend to have an isolating effect on those who don't know what they stand for. With `n-heroku-tools` you can ascertain that they are tools you can use with Heroku, but `nht` could be anything ('naughty hungry tortoises').

My feeling is that references in other repos should expand references from `nht` to `n-heroku-tools`, except when `nht` is used as a command, in which instance I am _guessing_ it can be traced to `n-heroku-tools` from its dependencies.

nickcolley: Let's switch it round...

nickcolley: I think best to use the common name if it's only used at `nht` but given the repo name isnt nht then this makes sense.

Also would prefer if we didnt use `nht` as well, since this is mostly used in make files, it could been expanded cross the board...

PR #605, opened 2020-08-11
Delete and rebuild review apps missing build

andygout:
Jira card: [Investigate Heroku review apps issues](https://financialtimes.atlassian.net/secure/RapidBoard.jspa?rapidView=1051&modal=detail&selectedIssue=CPP-159).

CircleCI builds are encountering failures that look like this: https://circleci.com/gh/Financial-Times/next-preflight/6729.

```
{ Error: No review app build found for app id '8e6b4116-2f77-438e-a35e-6f0a98b05d46';, commit '75b2f4e0a27ece478170d1d404c0cb347406d0df'
	at getAppBuildWithCommit.then (/home/circleci/project/build/node_modules/@financial-times/n-heroku-tools/lib/review-apps.js:161:12)
	at <anonymous>
	at process._tickDomainCallback (internal/process/next_tick.js:229:7) attemptNumber: 61, retriesLeft: 0 }
```

It usually (quite possibly exclusively) occurs following a rebase with another branch, though engineering a rebase does not always reliably reproduce the issue.

The solution developers have been using to fix this problem is to delete the Heroku review app and re-run the CircleCI workflow, which is a time-wasting process.

This PR implements logic so that in the event of an app ID not having a corresponding build, the review app that is no longer usable is deleted and the process to create a review app starts over.

The `review-apps.js` file (amended in this PR) uses a package called `pRetry`, which allows functions to (as the name suggests) retry a function for a specified number of attempts, with a specified interval between each. A normal error will trigger a retry of the function but an abortive error (`pRetry.AbortError`) will break out of the retry loop.

Such a retry loop has now been applied to the main function in this file (i.e. the one that covers the entire creation process) so that in the event of an absent build a retry can be attempted (post-deletion of the unusable review app).

This has introduced some complexity in that some of the functions this main function calls have their own retry loops, and so the logic depends on the correct type of error thrown in the right place, i.e. `Error` or `pRetry.AbortError`.

I have refactored the functions to use `await` exclusively for handling asynchronous code as there was a mix with `.then` and `.catch`, as well as renaming some of the functions and adding more unit tests.

Naturally, now that I want to test it on production (by pointing an app to this branch of `n-herouk-tools` as its dependency) I cannot reproduce the error, but thought it would be useful to get PR feedback in the meantime while I figure that out.

### DIFF
--- a/lib/review-apps.unit.test.js
+++ b/lib/review-apps.unit.test.js
@@ -2,33 +2,45 @@ require('isomorphic-fetch');
 const nock = require('nock');
 const pRetry = require('p-retry');
 
-jest.mock('./github-api');
 const reviewApps = require('./review-apps');
+
 const BASE_HEROKU_API_URL = 'https://api.heroku.com';
 
+jest.mock('./github-api');
+
 describe('review-apps', () => {
+
 	let nockScope;
 	let reviewAppsNockScope;
+
 	beforeAll(() => {
+
 		nockScope = nock(BASE_HEROKU_API_URL)
 			.matchHeader('accept', 'application/vnd.heroku+json; version=3');
+
 		reviewAppsNockScope = nock(BASE_HEROKU_API_URL)
 			.matchHeader('accept', 'application/vnd.heroku+json; version=3.review-apps');
+
 		jest.mock('shellpromise', () => a => a, { virtual: true });
 
 		// WARNING: Disable error logs, to clean up test output
 		// Re-enable here if need be
 		jest.spyOn(console, 'error').mockImplementation().mockName('console.error');
+
 	});
 
 	afterAll(() => {
+
 		jest.unmock('shellpromise');
 		jest.unmock('./github-api');
+
 	});
 
 	afterEach(() => {
+
 		nock.cleanAll();
 		jest.resetAllMocks();
+
 	});
 
 	describe('handleNonSucceededReviewAppBuild', () => {
@@ -88,9 +100,11 @@ describe('review-apps', () => {
 	});
 
 	describe('createReviewApp', () => {
+
 		const { createReviewApp } = reviewApps;
 
 		it('returns created review app', async () => {
+
 			const pipelineId = 'pipeline123';
 			const branch = 'some-branch';
 			const repoName = 'next-amazing';
@@ -100,13 +114,7 @@ describe('review-apps', () => {
 			reviewAppsNockScope.post('/review-apps')
 				.reply(200, {});
 
-			return createReviewApp({
-				pipelineId,
-				branch,
-				repoName,
-				commit,
-				githubToken
-			})
+			return createReviewApp({ pipelineId, branch, repoName, commit, githubToken })
 				.then(async res => {
 					const { ok } = res;
 					expect(ok).toBeTruthy();
@@ -114,9 +122,11 @@ describe('review-apps', () => {
 					const reviewApp = await res.json();
 					expect(reviewApp).toBeTruthy();
 				});
+
 		});
 
 		it('returns an error', async () => {
+
 			const pipelineId = 'pipeline123';
 			const branch = 'some-branch';
 			const repoName = 'next-amazing';
@@ -126,13 +136,7 @@ describe('review-apps', () => {
 			reviewAppsNockScope.post('/review-apps')
 				.reply(400, {});
 
-			return createReviewApp({
-				pipelineId,
-				branch,
-				repoName,
-				commit,
-				githubToken
-			})
+			return createReviewApp({ pipelineId, branch, repoName, commit, githubToken })
 				.then(async res => {
 					const { ok } = res;
 					expect(ok).toBeFalsy();
@@ -140,57 +144,61 @@ describe('review-apps', () => {
 					const error = await res.json();
 					expect(error).toBeTruthy();
 				});
+
 		});
+
 	});
 
 	describe('findCreatedReviewApp', () => {
+
 		const { findCreatedReviewApp } = reviewApps;
 
 		it('returns review app of branch', async () => {
 			const pipelineId = 'pipeline123';
 			const branch = 'some-branch';
+
 			reviewAppsNockScope.get('/pipelines/pipeline123/review-apps')
 				.reply(200, [
 					{
 						branch
 					}
 				]);
-			const reviewApp = await findCreatedReviewApp({
-				pipelineId,
-				branch
-			});
-			expect(reviewApp).toEqual({
-				branch
-			});
+
+			const reviewApp = await findCreatedReviewApp({ pipelineId, branch });
+			expect(reviewApp).toEqual({ branch });
+
 		});
 
 		it('returns undefined if branch review app is not found', async () => {
+
 			const pipelineId = 'pipeline123';
 			const branch = 'some-branch';
+
 			reviewAppsNockScope.get('/pipelines/pipeline123/review-apps')
 				.reply(200, []);
-			const reviewApp = await findCreatedReviewApp({
-				pipelineId,
-				branch
-			});
+
+			const reviewApp = await findCreatedReviewApp({ pipelineId, branch });
 			expect(reviewApp).toBeUndefined();
+
 		});
 
 		it('throws an error', async () => {
+
 			const pipelineId = 'pipeline123';
+
 			reviewAppsNockScope.get('/pipelines/pipeline123/review-apps')
 				.reply(400, {
 					message: 'Some error occurred'
 				});
 
 			try {
-				await findCreatedReviewApp({
-					pipelineId
-				});
+				await findCreatedReviewApp({ pipelineId });
 			} catch (error) {
 				expect(error.message).toEqual('Some error occurred');
 			}
+
 		});
+
 	});
 
 	describe('handleWaitTillReviewAppCreated', () => {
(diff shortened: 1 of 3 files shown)


andygout:
The `next-es-interface` CircleCI workflow was failing on [this branch](https://github.com/Financial-Times/next-es-interface/pull/1475) owing to the `No review app build found` error : https://circleci.com/gh/Financial-Times/next-es-interface/10544.

I added a commit to that branch so that its `n-heroku-tools` dependency pointed to this branch, and it fixed the issue by deleting the unusable review app and created a new one: https://circleci.com/gh/Financial-Times/next-es-interface/10547.

N.B. I have since removed the commit which is why you won't see it in the commit history of that branch.

andygout:
In terms of a release, even thought it's a lot of code changes, it's still a patch, i.e. fixes a bug without adding any new functionality.

Do others agree?

lib/review-apps.unit.test.js
```diff
@@ -1,39 +1,110 @@
...
 		nock.cleanAll();
 		jest.resetAllMocks();
+
+	});
+
+	describe('runCheckForBuildSuccessStatus', () => {
+
+		const { runCheckForBuildSuccessStatus } = reviewApps;
+
+		it('returns without throwing an error if awaiting the review app build does not error', async () => {
```
antonsamper, line 50: Can this test be removed? The response is mocked so in essence its testing `nock` works correctly

antonsamper:
As a general note, I've notice the line breaks. Does the FT or do we as a team have guidance/a preference on this? I'm just thinking for consistency.

lib/review-apps.unit.test.js
```diff
@@ -1,39 +1,110 @@
...
+
+		});
+
 	});
 
 	describe('createReviewApp', () => {
+
 		const { createReviewApp } = reviewApps;
 
 		it('returns created review app', async () => {
```
antonsamper, line 106: I know you didn't write this test but i don't think its a good test in its current state.

It could be improved by doing the following:
- asserting the body of the request
- adding some data in the response to assert the full response is returned from `createReviewApp`

lib/review-apps.unit.test.js
```diff
@@ -43,23 +114,19 @@ describe('review-apps', () => {
...
+			const { ok } = response;
+
+			const reviewApp = await response.json();
+
+			expect(ok).toBeTruthy();
+			expect(reviewApp).toBeTruthy();
+
 		});
 
 		it('returns an error', async () => {
```
antonsamper, line 128: This test is the same as `returns created review app`

lib/review-apps.unit.test.js
```diff
@@ -217,183 +406,111 @@ describe('review-apps', () => {
...
+				await repeatedCheckForBuildSuccessStatus({ commit, appId });
+			} catch (error) {
+				expect(true).toBeFalsy();
+			}
 
-			const waitedReviewAppBuild = await waitForReviewAppBuild({
-				commit,
-				minTimeout: 0
-			})(appId);
+			expect(true).toBeTruthy();
```
antonsamper, line 454: This will always pass

lib/review-apps.unit.test.js
```diff
@@ -217,183 +406,111 @@ describe('review-apps', () => {
...
-				minTimeout: 0
-			})(appId);
+			nockScope.get('/apps/app123/builds')
+				.times(60)
+				.reply(400, {
+					message: 'Some error occurred'
+				});
+
+			try {
+				await repeatedCheckForBuildSuccessStatus({ commit, appId, minTimeout: 0 });
```
antonsamper, line 470: Is there a way to check how many times the mock has been called? I know jest has https://jestjs.io/docs/en/expect#tohavebeencalledtimesnumber

lib/review-apps.unit.test.js
```diff
@@ -217,183 +406,111 @@ describe('review-apps', () => {
...
-				await waitForReviewAppBuild({
-					commit,
-					minTimeout: 0
-				})(appId);
-			} catch (e) {
-				const { message } = e;
-				expect(message).toMatch('Review app build failed');
-				expect(message).toMatch('https://heroku.com/builds/output/app123');
+				await repeatedCheckForBuildSuccessStatus({ commit, appId, minTimeout: 0 });
+			} catch (error) {
```
antonsamper, line 495: Similar to above, is there a way to check how many times the mock has been called? I know jest has https://jestjs.io/docs/en/expect#tohavebeencalledtimesnumber

lib/review-apps.unit.test.js
andygout: The reasons I added it were:
- There is a corresponding test for when an error is thrown to ensure that it in that scenario it throws a specific type of error (`pRetry.AbortError`), so seemed logical to have a the flip-side of that scenario (i.e. no error thrown). I wanted to achieve coverage of scenarios as comprehensive as possible so that they were all accounted for (which inevitably results in a few tests with minimal value such as this one).
- There is (admittedly minimal) value in knowing that it simply returns once it has finished its async business without returning a specific value (though admittedly this is not explicit in the test).
- This function also relies on the success of the other functions called before it encounters `nock`: `repeatedCheckForBuildSuccessStatus()`, `checkForBuildSuccessStatus()`, `getAppBuildWithCommit()`, `getBuilds()`, `getBuildsUrl()`, which is inadvertently also being tested.

But I take your point and am happy to remove if you still think it does not provide sufficient value?

lib/review-apps.unit.test.js
andygout: Good spot! I was only editing this test to use `await` (i.e. not `.then()`) and did not notice that it is not actually testing for an error to be returned (also unclear if this means an error should be returned or thrown).

I will amend it to test what is in its description.

lib/review-apps.unit.test.js
andygout: I added console logs just before `expect(true).toBeFalsy();` and `expect(true).toBeTruthy();` respectively.

If an error is thrown by `repeatedCheckForBuildSuccessStatus()` then it is caught and the expectation that always fails will run and so cause the test to fail. In this scenario, `expect(true).toBeTruthy();` is not reached, and even if it were, the expectation that always fails will also have been encountered and cause the test to fail overall.

```js
try {
	await repeatedCheckForBuildSuccessStatus({ commit, appId });
} catch (error) {
	expect(true).toBeFalsy();
}

expect(true).toBeTruthy();
```

It's not ideal as it's not particularly explicit as to the intention of the test, but I could not get this working with `expect(() => ...error...).not.toThrow(error)` when accommodating an async function (re. https://stackoverflow.com/questions/49603338/how-to-test-an-exception-was-not-thrown-with-jest#answer-49603347). I think this would be a preferable solution if we can get it working.

lib/review-apps.unit.test.js
andygout: > asserting the body of the request

Do you mean the body of the request made to `nock` (in this instance the POST request made to `/review-apps`)?

> adding some data in the response to assert the full response is returned from `createReviewApp`

👍

antonsamper: > Do you mean the body of the request made to `nock`?

yes, that way we would know that the correct request is being made

lib/review-apps.unit.test.js
antonsamper: I see your thinking behind. Lets not remove it but maybe we can tweak the assertion to checking that it doesnt error?

lib/review-apps.unit.test.js
andygout: Having looked at this test and the corresponding function, while ostensibly it only tests that Nock is working, I think it is important to test that this function does not throw an error in the instance of the API returning a 400.

There is a function in `review-apps.js` - `performFetch` - that all other data fetches use, and which will throw an error if the `ok` value in the response is false.

It is especially important for this specific function that it does not use `performFetch` so that an error is not thrown (as there is logic that needs to be performed in the event of a non-OK response), and so I think that justifies keeping it (though I have removed the expectation where it expects `error` from this line (which is just an empty object `{}`) to be truthy: `const error = await response.json();`).

lib/review-apps.unit.test.js
andygout: I have used `nock.isDone()`, taken from this blog post: [Ensure All Nock Interceptors Are Used by neverendingqs](https://blog.neverendingqs.com/2017/03/13/ensure-all-nock-interceptors-are-used.html).

lib/review-apps.unit.test.js
andygout: I have used `nock.isDone()`, taken from this blog post: [Ensure All Nock Interceptors Are Used by neverendingqs](https://blog.neverendingqs.com/2017/03/13/ensure-all-nock-interceptors-are-used.html).

nickcolley:
> In terms of a release, even thought it's a lot of code changes, it's still a patch, i.e. fixes a bug without adding any new functionality.

Do others agree?

If it does not break existing builds, and does not expose a feature as part of it's public api then yes it's a patch.

antonsamper:
> > In terms of a release, even thought it's a lot of code changes, it's still a patch, i.e. fixes a bug without adding any new functionality.
> 
> Do others agree?
> 
> If it does not break existing builds, and does not expose a feature as part of the public api then yes it's a patch.

Yep, it should be a patch

andygout:
Closes https://github.com/Financial-Times/n-heroku-tools/issues/570.

andygout:
Fixes https://github.com/Financial-Times/n-heroku-tools/issues/570

PR #606, opened 2020-08-21
CPP-249 -  unclear errors thrown by `nht configure`

emortong:
[ticket](https://financialtimes.atlassian.net/browse/CPP-249)

- ~~removes config variables if their values are empty string or number~~
- if a variable is invalid it logs a warning message
- it adds a hint to the error saying to review Vault for invalid variables.

### DIFF
--- a/tasks/configure.js
+++ b/tasks/configure.js
@@ -92,17 +92,12 @@ const getPipelineId = async (pipelineName) => {
 	return pipeline.id;
 };
 
-const validateVariables = (obj) => {
-	// removes empty strings and numbers
-	let filteredObj = {};
+const warnForInvalidVariables = (obj) => {
 	Object.keys(obj).forEach(prop => {
-		if (obj[prop] !== '' && typeof obj[prop] !== 'number') {
-			filteredObj[prop] = obj[prop];
-		} else {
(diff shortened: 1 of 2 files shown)


nickcolley (approved):
Is looking, would be good to get @apaleslimghost thoughts too if possible but 👍 from me

tasks/configure.js
```diff
@@ -152,6 +152,11 @@ async function task (opts) {
 		}
 	});
 
+	const invalidVariables = Object.keys(patch).filter(prop => patch[prop] === '' || typeof patch[prop] === 'number');
```
nickcolley, line 155: If it is possible to add a test for this that'd be great, but I'm not sure if this repository has tests.

apaleslimghost (approved):
yeah, this is good. with filtering/coercing bad values i did suggest logging a warning, but that could be easily overlooked especially in CI, and it would be confusing and potentially dangerous for apps to not get given the same values you put in Vault so 👍

tasks/configure.js
apaleslimghost: there is a test for this task (https://github.com/Financial-Times/n-heroku-tools/blob/225fc1bac437dce61928f41b60b06291edf50a4b/tasks/configure.test.js), adding this would be nice but currently it looks like it's not covering any error handling at all so it might be too much effort to be worth it

PR #607, opened 2020-08-26
chore(upload-assets-to-s3): upload svg files

antonsamper:
#### Reference
- https://financialtimes.atlassian.net/browse/CPP-230

#### Notes
- Add `svg` to list of extensions to upload

### DIFF
--- a/tasks/upload-assets-to-s3.js
+++ b/tasks/upload-assets-to-s3.js
@@ -7,7 +7,7 @@ const aws = require('aws-sdk');
 const defaultDirectory = 'public';
 const defaultBucket = 'ft-next-hashed-assets-prod';
 const defaultDestination = 'hashed-assets/uploaded';
-const defaultFileExtensions = ['js', 'css', 'map', 'gz', 'br', 'png', 'jpg', 'jpeg', 'gif', 'webp'].join();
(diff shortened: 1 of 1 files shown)


nickcolley (approved):
Makes sense 👍

tasks/upload-assets-to-s3.js
```diff
@@ -7,7 +7,7 @@ const aws = require('aws-sdk');
 const defaultDirectory = 'public';
 const defaultBucket = 'ft-next-hashed-assets-prod';
 const defaultDestination = 'hashed-assets/uploaded';
-const defaultFileExtensions = ['js', 'css', 'map', 'gz', 'br', 'png', 'jpg', 'jpeg', 'gif', 'webp'].join();
```
nickcolley, line 10: since these are the defaults can also consider extending them if it's only required in one place.

apaleslimghost: this command is called by `n-gage` which doesn't currently offer a way to set this argument, so the app would have to override that task in its makefile to add the argument, which makes me queasy

i think SVG would be a common enough use case that it should be a default?

antonsamper: Yeah that makes sense but it would be a bit more involved, something to consider next time

nickcolley: > this command is called by n-gage which doesn't currently offer a way to set this argument

Didn't realize that, makes sense to me.

PR #629, opened 2021-06-24
Delete rebuild command

emortong:
It was only being used by `n-ui` and `n-ui` had been deprecated and archived.


apaleslimghost (approved):
nice! 🔥 🗑️

PR #661, opened 2023-03-21
Uppercase all HTTP verbs for native Fetch support

olusoladavid:
### The problem
Standard Fetch spec will only automatically uppercase RFC9110-compliant http verbs - see [MDN docs](https://developer.mozilla.org/en-US/docs/Web/API/fetch). This implies that custom methods like `PATCH` will be broken if they are passed to the Fetch API as lowercase.

This becomes an issue for n-gage heroku projects migrating to Node 18 which include a native global `fetch` and replaces `isomorphic-fetch` polyfill. CI workflows to deploy  review apps will fail as seen [here](https://app.circleci.com/pipelines/github/Financial-Times/next-consent-proxy/2210/workflows/52544f57-a8ad-4824-8d23-d9f0138a2d81) because the HTTP verb `patch` is invalid for configuring the Heroku pipeline

### Notes
See this [Slack conversation](https://financialtimes.slack.com/archives/C3TJ6KXEU/p1679317662364499) for context

### Solution
This commit ONLY sets a precedent/convention for using uppercase methods in the project as there's no centralized callsite for `fetch` in the project and no easy way to override the behavior for native fetch. Also, projects are expected to migrate away from `n-heroku-tools` in the nearest future

### DIFF
--- a/lib/heroku-config-vars.js
+++ b/lib/heroku-config-vars.js
@@ -49,7 +49,7 @@ class HerokuConfigVars {
 	*/
 	set (patch) {
 		return this.makeApiCall({
-			method: 'patch',
+			method: 'PATCH',
 			body: JSON.stringify(patch)
 		});
 	}
--- a/lib/review-apps.js
+++ b/lib/review-apps.js
@@ -125,7 +125,7 @@ const deleteReviewAppAndThrowError = async ({ reviewApp, commit, appId }) => {
 
 		const headers = await herokuHeaders({ useReviewAppApi: true });
 
-		await performFetch(getReviewAppUrl(reviewApp.id), { headers, method: 'delete' });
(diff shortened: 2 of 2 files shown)


ivomurrell (approved):
Even though n-heroku-tools is unsupported, and we haven't worked out how to support global `fetch` in Platforms' repositories yet, I think this is still worth merging this fix. Seeing as n-heroku-tools uses `isomorphic-fetch` rather than `node-fetch` directly it creates an expectation that it should be compatible with the web API's `fetch`, which it isn't currently. I think it's okay to make a quick fix for that now and then consider whether we actually want to support a global `fetch` or whether we should have been using `node-fetch` directly later on.

Anyway, thanks for all your work looking into this with me, Sola!

